Issue #62, opened 2019-08-23
Can't create an HP inside of a Layer's `call`


omalleyt12:
This gives an error in deserialize_keras_object about not recognizing `Float` (likely something to do with how `module_objects` are being picked up):

```python
from tensorflow import keras
import kerastuner
import numpy as np

def build_model(hp):
    class Bias(keras.layers.Layer):
        def call(self, x):
            return x + hp.Float('bias', 0, 2, step=0.5)

    layer = Bias(input_shape=(1,))
    model = keras.Sequential([layer])
    model.compile('sgd', 'mse', metrics=['accuracy'])
    return model

x = np.zeros((10, 1))
y = np.ones((10, 1))

rs = kerastuner.tuners.RandomSearch(build_model, 'val_accuracy', 10)
rs.search(x, y, epochs=1, validation_data=(x, y))
```

Issue #54, opened 2019-08-13
Weights for model sequential have not yet been created error


daviddexter:
I'm getting a weights error when run a simple tuner.

ENV : tf version 22.0.0-beta1
         keras 2.2.4-tf

```py
class FBPDHyperModel(HyperModel):
    def __init__(self,features):
        self.features_layers = features # DenseFeatures Layer         

    def build(self, hp):
        model = tf.keras.Sequential()
        model.add(self.features_layers)
        model.add(layers.Dense(units=hp.Range('units',32, 512, 32),activation='relu'))
        model.add(layers.Dense(3, activation='softmax'))
        model.compile(optimizer=tf.keras.optimizers.Adam(),loss='sparse_categorical_crossentropy',metrics=['accuracy'])
        return model  

tuner = RandomSearch(
    FBPDHyperModel(features=features_layer),
    objective='val_accuracy',
    max_trials=5,
    directory='tuner_dir')
```

On running, I get this error

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-49-720dad76188f> in <module>
      3     objective='val_accuracy',
      4     max_trials=5,
----> 5     directory='tuner_dir')

/opt/conda/lib/python3.6/site-packages/kerastuner/tuners/randomsearch.py in __init__(self, hypermodel, objective, max_trials, seed, **kwargs)
    120             objective,
    121             max_trials,
--> 122             **kwargs)

/opt/conda/lib/python3.6/site-packages/kerastuner/engine/tuner.py in __init__(self, oracle, hypermodel, objective, max_trials, executions_per_trial, max_model_size, optimizer, loss, metrics, hyperparameters, tune_new_entries, allow_new_entries, distribution_strategy, directory, project_name)
    200         # Populate initial search space
    201         if not self.hyperparameters.space and self.tune_new_entries:
--> 202             self._build_model(self.hyperparameters)
    203 
    204     def search(self, *fit_args, **fit_kwargs):

/opt/conda/lib/python3.6/site-packages/kerastuner/engine/tuner.py in _build_model(self, hp)
    599 
    600             # Check model size.
--> 601             size = utils.compute_model_size(model)
    602             if self.max_model_size and size > self.max_model_size:
    603                 oversized_streak += 1

/opt/conda/lib/python3.6/site-packages/kerastuner/utils.py in compute_model_size(model)
     21 def compute_model_size(model):
     22     "comput the size of a given model"
[...]

PR #64, opened 2019-08-23
Add support for Subclass Models and Sequential Models w/o input_shape

omalleyt12:
* Don't compute the model size if the model is not built

* Change deserialization module_objects to support autograph'd layers


omalleyt12:
Fixes:

https://github.com/keras-team/keras-tuner/issues/62
https://github.com/keras-team/keras-tuner/issues/54

PR #77, opened 2019-09-11
Make hyperband oracle obey objective direction.

### LINKED ISSUE #72: hyperband doesn't work as intended
There sorting of the trial is incorrect for at least two reasons:
1. line: 130 the values are not initialized so the comparison fail in line 136
2. The sort in 136: sort in ascending order so the one with the lowest score are always picked up in line 139

So the init/sort should take into account the metric direction

jamlong:
Fixes #72.

The existing implementation of hyperband did not take into account the objective direction when sorting candidates during selection.  The older implementation inverted the scores when necessary (which was rather ugly) and that bit appears to have been lost in the refactor. Thus, when a max-sorted objective was used, HyperbandOracle._select_candidates would sort in the wrong direction.

HyperbandTuner takes the objective name, and uses the infer_metric_direction_from_name introduced in https://github.com/keras-team/keras-tuner/pull/76, then passes the resulting direction as a constructor argument to the Oracle. 

Added a unit test showing that max, min, and default (==min) sorting all now work.

### DIFF
--- a/kerastuner/tuners/hyperband.py
+++ b/kerastuner/tuners/hyperband.py
@@ -16,7 +16,9 @@ import queue
 import random
 import json
 from ..engine import tuner as tuner_module
+from ..engine import tuner as tuner_module
 from ..engine import oracle as oracle_module
+from ..engine.metrics_tracking import infer_metric_direction_by_name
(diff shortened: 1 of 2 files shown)


jamlong:
Note: must be merged after https://github.com/keras-team/keras-tuner/pull/76

PR #81, opened 2019-09-12
Fixes #74 - handle val_accuracy (and similar) as objectives

### LINKED ISSUE #74: Confusing (and incorrect) results_summary and error for weight initilisation
Hi all, 

I was wondering if someone could clarify a few things for me, i would be very greatful! :)

I have run keras tuner with the following code in order to optimise a model by unit number and layer number: 

```
def build_model(hp):
    model = keras.Sequential()
    for i in range(hp.Int('num_layers', 2, 20)):
        model.add(layers.Dense(units=hp.Int('units_' + str(i),
                                            min_value=32,
                                            max_value=512,
                                            step=32),
                               activation='relu'))
    model.add(layers.Dense(24, activation='sigmoid'))
    model.compile(
        optimizer=keras.optimizers.Adam(
            hp.Choice('learning_rate', [1e-2, 1e-3, 1e-4])),
        loss='binary_crossentropy',
        metrics=['accuracy'])
    return model

tuner = RandomSearch(
    build_model,
    objective='val_accuracy',
    max_trials=5,
    executions_per_trial=5,
    directory='/Users/blablabla/Desktop/',
    project_name='Optimal model')

tuner.search_space_summary()

tuner.search(X_train.values, y_train,
             epochs=50,
             validation_data=(X_test.values, y_test),
             shuffle=True
             )
```

This code runs **without** any issue! however.. when calling `tuner.results_summary()`

The results summary are completely of the actual training results (both during hyperparameter training but also if i just train the model without the hyperparamater (normal keras))

```
tuner.results_summary=
[Results summary]
 |-Results in /Users/jassim01/Desktop/CRUK Cambridge/R/Amir/Optimal model
 |-Ran 5 trials
 |-Ran 25 executions (5 per trial)
 |-Best val_accuracy: **0.5650**
```

the best val_acc is 0.56 and this is completely wrong according to the best performance i can achieve when running keras alone, or during the hyperparameter training. during hyperparameter training I am achieving (according to hyperparameter tuning progress report in excess of 90%[see one line as example]) 

is this is a bug or am i misintepreting something?

[CPU: 43%]Epoch 28/50: 100%|██████████| 5/5 [00:00<00:00, 61.46steps/s, loss=0.09, accuracy=0.969, val_loss=0.184, val_accuracy=0.913

Secondly, when calling best_model:

best_model = tuner.get_best_models(num_models=1)[0]

I get the error: ValueError: Weights for model sequential_2 have not yet been created. Weights are created when the Model is first called on inputs or `build()` 

I understand this error, however none of the tutorials call model.build argument at the end of the for loop in the examples. Since the model is actually able to train, is this a bug? 

if not.. is it ok to add an **input_dim** inside the for loop as follows? 

model.add(layers.Dense(**input_dim=5078**, units=hp.Int('units_' + str(i),
                                            min_value=32,
                                            max_value=512,
                                            step=32),.. etc

If i do this, tuner.results_summary()
shows a best val_acc of 0.90... and get_bestmodel also works.... 

just want to confirm this is correct.. 

Thanks for your help!!

jamlong:
metrics_tracking.py relied on class names when determining directions. However, this misses the use case of the objective - if the objective is val_accuracy (or val_anything_that_should_sort_max), when you call tuner.get_best_models() or similar functions, the metrics_tracking.MetricsTracker used will NOT have other metrics, which means the prior handling of objective won't work.

Previously, it would check if name[:4] is known to have a direction. But in the case of objective=val_accuracy, the metric accuracy isn't necessarily registered.

To handle this, I've added infer_metric_direction_by_name() to use in the objective case.
infer_metric_direction() already exists, but operates on a Metrics object.

I can additionally change infer_metric_direction() to use infer_metric_direction_by_name() if desired, although it offers the slim chance of a conflict on the, currently separate, _MAX_METRICS and _MAX_METRICS_FN.

### DIFF
--- a/kerastuner/engine/metrics_tracking.py
+++ b/kerastuner/engine/metrics_tracking.py
@@ -21,7 +21,6 @@ import numpy as np
 
 
 class MetricsTracker(object):
-
     def __init__(self, metrics=None):
         self.names = []
         self.directions = {}
@@ -38,19 +37,16 @@ class MetricsTracker(object):
             self.register(metric.name, direction)
 
     def register(self, name, direction=None):
+        """Register a metric by name and direction."""
+
         if direction is None:
-            if name.startswith('val_') and name[4:] in self.names:
-                direction = self.directions[name[4:]]
-            else:
-                # Assume `min` by default
-                direction = 'min'
-                # TODO: raise warning
+            direction = infer_metric_direction_by_name(name)
+
         if direction not in {'min', 'max'}:
-            raise ValueError(
-                '`direction` should be one of '
-                '{"min", "max"}, but got: %s' % (direction,))
+            raise ValueError('`direction` should be one of '
+                             '{"min", "max"}, but got: %s' % (direction, ))
         if name in self.names:
-            raise ValueError('Metric already exists: %s' % (name,))
+            raise ValueError('Metric already exists: %s' % (name, ))
         self.names.append(name)
         self.directions[name] = direction
         self.metrics_history[name] = []
@@ -73,7 +69,7 @@ class MetricsTracker(object):
 
     def get_history(self, name):
         if name not in self.names:
-            raise ValueError('Unknown metric: %s' % (name,))
+            raise ValueError('Unknown metric: %s' % (name, ))
         return self.metrics_history[name]
 
     def set_history(self, name, series):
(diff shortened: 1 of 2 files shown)


ebursztein:
Can you check why the test fails? we can't merge something with failing tests

omalleyt12:
@jamlong this should be fixed in the refactor, could you please give it a try?

omalleyt12:
Going to close this out, please reopen if you find there is still an issue

PR #92, opened 2019-09-27
Add protos for HyperParameter classes

omalleyt12:
Also adds a `to_proto` and `from_proto` for each class


omalleyt12:
@sagipe

.gitignore
```diff
@@ -16,4 +16,4 @@ tmp.py
 .coverage
 .coverage.*
 htmlcov
-old_code/
\ No newline at end of file
+old_code/
```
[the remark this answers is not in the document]
omalleyt12: Acknowledged, I think this was just picked up because I was originally trying to exclude the *_pb2.py files from git history, but decided to include them so that external devs don't need to compile protos

omalleyt12:
> A meta comment: for Tuner/Oracle communication and state serialization, we should prefer JSON when possible

SG, I think we'll need these protos anyway in order to integrate with TFX, Cloud, TensorBoard etc.

For state serialization I think saving as JSON makes sense. My worry with using JSON for Tuner/Oracle communication is that it may make it harder for us to integrate with various cloud services like CMLE and OSS TFX. I could envision cases where they want to run an OracleService to get HyperParameter protos and then pass those to downstream processes

kerastuner/protos/kerastuner.proto
```diff
@@ -0,0 +1,88 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+syntax = "proto3";
+
+package kerastuner;
```
sagipe: Since the main use case for this would be for internal integration, we should consider the proper format for these protos.

We already have a representation of HParams for Tensorboard at:
https://github.com/tensorflow/tensorboard/blob/2b5547fe5fb96eba08f14ed2a97b2f8d519423e3/tensorboard/plugins/hparams/api.proto
and example parsing code in:
https://github.com/tensorflow/tensorboard/blob/2b5547fe5fb96eba08f14ed2a97b2f8d519423e3/tensorboard/plugins/hparams/summary_v2.py

We should consider how specialized this proto should be for KerasTuner VS if this should be a more generic proto that can be used in Tensorboard and other config files; and the KerasTuner HyperParameters object can initialize from it. I'll follow up offline.

kerastuner/protos/kerastuner.proto
```diff
@@ -0,0 +1,88 @@
...
+
+    bool ordered = 8;
+}
+
+message Boolean {
+    string name = 1;
+    bool default = 2;
+}
+
+message HyperParameters {
```
sagipe: Here and elsewhere, google.protobuf.Value may be a better choice, rather than implicit types.

PR #99, opened 2019-10-02
Removed all unused variables and enabled check.

gabrieldemarmiesse:


omalleyt12:
@gabrieldemarmiesse could you update to fix the other new flake8 rules? Thanks!

gabrieldemarmiesse:
@omalleyt12 could you use "Squash and merge" when merging from now on? The history of the master branch is then much cleaner and easier to follow. Plus, since all commits are passing the CI, we can make use of tools like git-biscect, which saves lives! We used this merge option in the keras repository and it was a really nice experience :)

omalleyt12:
@gabrieldemarmiesse Sure will do :)

PR #102, opened 2019-10-04
Implement the rest of the GRPC methods and add an OracleClient

omalleyt12:
The oracle client will be used to wrap and replace an `oracle` in `BaseTuner.__init__`

```python
self.oracle = OracleClient(oracle)
```

Also, found an error in how I was mocking env vars, so changed the implementation and added more testing


omalleyt12:
Thanks for the review!

PR #109, opened 2019-10-08
Enabled other flake8 rules.

gabrieldemarmiesse:
After #103 was merged, we don't need to break as many pep8 rules as before.


omalleyt12:
LGTM awesome, thanks!

PR #112, opened 2019-10-09
Support Distributed Tuning

omalleyt12:
* Automatically starts a chief oracle if `KERASTUNER_ORACLE_IP` is set and `KERASTUNER_TUNER_ID == 'chief'`
* Wraps the Oracle in an `OracleClient` that makes call to the chief Oracle when `KERASTUNER_ORACLE_IP` is set and `KERATUNER_TUNER_ID != 'chief'`
* Adds test with 3 workers for BaseTuner and 1 worker for RandomSearch (TF models aren't thread-safe, so 1 worker and 1 chief was easier to test)


omalleyt12:
Thanks for the review!

PR #113, opened 2019-10-10
Fix MultiExecutionTuner to not copy hyperparameters in order for update_space to work

omalleyt12:
* Update_space was working for the initial model build
* But because we were copying HPs before sending to `_build_model`, the update_space that was called in `on_trial_end` was not seeing the new HyperParameters


tests/kerastuner/tuners/randomsearch_test.py
```diff
@@ -11,3 +11,52 @@
...
+from kerastuner.engine import multi_execution_tuner
+from kerastuner.tuners import randomsearch
+
+
+@pytest.fixture(scope='function')
+def tmp_dir(tmpdir_factory):
+    return tmpdir_factory.mktemp('randomsearch_test', numbered=True)
+
+
+def test_update_space(tmp_dir):
```
fchollet: Please add a comment describing the behavior that the test is checking.

kerastuner/engine/multi_execution_tuner.py
```diff
@@ -83,7 +83,7 @@ def run_trial(self, trial, *fit_args, **fit_kwargs):
         # Run the training process multiple times.
         metrics = collections.defaultdict(list)
         for execution in range(self.executions_per_trial):
-            model = self._build_model(trial.hyperparameters.copy())
+            model = self._build_model(trial.hyperparameters)
```
fchollet: Isn't it strange that `trial.hyperparameters` would be out of sync with `oracle.hyperparameters`?

omalleyt12: The only time these would be out of sync is if `allow_new_entries=True` and `tune_new_entries=False`. To me, it makes sense in that case to have the Trial show that it is using the default values for these HPs

This implementation also allows us to take care of `oracle.update_space` for users, by putting it in the `on_trial_end` hook. This way any HPs added during `run_trial` (even HPs for things like preprocessing) will be picked up by the Oracle

tests/kerastuner/tuners/randomsearch_test.py
omalleyt12: Done

haifeng-jin:
@Mohammedelkhalil thanks for the comment!

I do not see any bug as I reproduce the results with your snippet.
https://colab.research.google.com/github/haifeng-jin/Colabs/blob/main/KerasTuner_issue_113.ipynb

Would you please elaborate more on what is the expected output?

TyouzichaT:
> @Mohammedelkhalil thanks for the comment!
> 
> I do not see any bug as I reproduce the results with your snippet. https://colab.research.google.com/github/haifeng-jin/Colabs/blob/main/KerasTuner_issue_113.ipynb
> 
> Would you please elaborate more on what is the expected output?

During the the first trial, not all hyperparameters can be shown in the console.

image

From the screenshot above, you can see 'layers' is 7, but only 'unit0' was displayed.

PR #127, opened 2019-10-22
Remove conditional HPs from HyperResnet

omalleyt12:
Looking at this again the different values of the conv sizes for "next" doesn't seem to be a big enough distinction to warrant the additional complexity of having conditional parameters here


omalleyt12:
Thanks for the review!

PR #135, opened 2019-10-27
Bump keras autodoc version, with aliases now!

gabrieldemarmiesse:
To use aliases, use a string instead of the function/method/class. I've added an example for oracles.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ install:
   - travis_retry conda install portpicker
 
   # install tools for building the docs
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
It seems there are some flaky tests.

omalleyt12:
@gabrieldemarmiesse Awesome, thanks! Yeah there are some flaky tests right now, I'll send a PR out later to fix the flakes. For now rerunning

PR #137, opened 2019-10-28
Reduce flakiness

omalleyt12:
* max_value included in Float range
* Sklearn test too optimistic about generalization to the test set
* BayesianOptimization test too optimistic about finding best log-sampled 'c'


omalleyt12:
Thanks for the review!

PR #152, opened 2019-11-04
Fix Bayesian Choice boundary

omalleyt12:
When the `BayesianOptimization` oracle was hitting the `1` cumulative probabilty bound for `Choice`, `math.floor` was not properly rounding down resulting in an out of index error


omalleyt12:
Thanks for the review!

PR #156, opened 2019-11-06
Generate samples based on seed and update gaussian kernel

themrzmaster:
Generate random samples based on inpu seed

### DIFF
--- a/kerastuner/tuners/bayesian.py
+++ b/kerastuner/tuners/bayesian.py
@@ -116,7 +116,7 @@ class BayesianOptimizationOracle(oracle_module.Oracle):
         optimal_x = None
         num_restarts = 50
         bounds = self._get_hp_bounds()
-        x_seeds = self._random_state.uniform(bounds[:, 0], bounds[:, 1],
+        x_seeds = self._random_state.uniform(bounds[:, 0], bounds[:, 1], 
                                              size=(num_restarts, bounds.shape[0]))
         for x_try in x_seeds:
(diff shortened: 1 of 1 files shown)


omalleyt12:
Closed / reopened for build flake

omalleyt12:
@themrzmaster Can you please add a comment here when the PR is ready for review again? Dropping approval for now so I don't accidentally merge while you are working on it

themrzmaster:
It's ready! sorry for the many commits.

omalleyt12:
@themrzmaster Great thanks, no worries! Could you explain the use of `nu=2.5` in the Matern kernel? It looks like it is optimizing well, I am just curious for the reason for this change. Do you expect this to perform better than the default setting for the general case?

themrzmaster:
As proposed in http://papers.nips.cc/paper/4522-practical-bayesian-optimization-of-machine-learning-algorithms.pdf , the matern kernel with a 2.5 nu value results in sample functions which are twice-differentiable, which works better for quasi newton optimization algorithms as the one used (L-BFGS)

omalleyt12:
> @omalleyt12 As proposed in http://papers.nips.cc/paper/4522-practical-bayesian-optimization-of-machine-learning-algorithms.pdf , the matern kernel with a 2.5 nu value results in sample functions which are twice-differentiable, which works better for quasi newton optimization algorithms as the one used (L-BFGS)

Great, sounds good to me! Thanks for looking into this, merging now

PR #158, opened 2019-11-07
Expected Improvement acquisition function for bayesian optimization

themrzmaster:
Optional EI as acquisition function for Bayesian tuner.

### DIFF
--- a/kerastuner/engine/tuner_utils.py
+++ b/kerastuner/engine/tuner_utils.py
@@ -135,7 +135,7 @@ class UtilityFunction(object):
     @staticmethod
     def _ucb(x, gpr, kappa):
         mean, std = gpr.predict(x, return_std=True)
-        return mean + kappa * std
+        return mean - kappa * std
 
     @staticmethod
     def _ei(x, gpr, y_max, xi):
--- a/kerastuner/engine/tuner_utils.py
+++ b/kerastuner/engine/tuner_utils.py
@@ -20,6 +20,7 @@ from __future__ import print_function
 import math
 import numpy as np
 import six
+from scipy.stats import norm
 
 import tensorflow as tf
 from tensorflow import keras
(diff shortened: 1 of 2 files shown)


themrzmaster:
This paper https://arxiv.org/pdf/1206.2944.pdf provides a comparison between different acquisition functions, In that context, EI seems to perform better than UCB. I will look for more benchmarks

omalleyt12:
@themrzmaster Ok, I think it's probably good for us to add this argument

We should always be careful when adding new arguments, as these add cognitive overhead to our APIs for the user. In general, we expect users of Keras Tuner to be machine learning practitioners, who  probably don't have much experience with bayesian optimization. As such, we should make sure to provide our users with clear guidance

In this case I think it is reasonable to allow different acquisition functions, I'll add comments to the PR

kerastuner/engine/tuner_utils.py
```diff
@@ -116,3 +117,28 @@ def average_histories(histories):
     # Convert {str: [float]} to [{str: float}]
     averaged = [dict(zip(metrics, vals)) for vals in zip(*averaged.values())]
     return averaged
+
+
+class UtilityFunction(object):
```
omalleyt12: Since this is specific to BayesianOptimization, this should live in ```kerastuner.tuners.bayesian.py```

I think we should make this into a class hierarchy, something like:

```python
class AcquisitionFunction(object):
  def __call__(self, x, gpr, y_max):
    raise NotImplementedError

class UCB(AcquisitionFunction):
  ...

class EI(AcquisitionFunction):
  ...
```

omalleyt12: This allows us to support users providing a subclass of this class to the BayesianOracle.

We also need to make sure we are sure what the signature of ```__call__``` should be

kerastuner/tuners/bayesian.py
```diff
@@ -32,6 +33,8 @@ class BayesianOptimizationOracle(oracle_module.Oracle):
             as initial training data for Bayesian optimization. If not specified,
             a value of 3 times the dimensionality of the hyperparameter space is
             used.
+        acq: String. Acquisition function. 'ucb' for Upper Confidence Bound
```
omalleyt12: I think it's better to name this `acquisition_function`. We should accept the strings "ucb" and "ei", and also accept any subclass of ```AcquisitionFunction```

kerastuner/tuners/bayesian.py
```diff
@@ -107,21 +112,19 @@ def _populate_space(self, trial_id):
...
 
-        def _upper_confidence_bound(x):
-            x = x.reshape(1, -1)
-            mu, sigma = self.gpr.predict(x, return_std=True)
-            return mu - self.beta * sigma
-
         optimal_val = float('inf')
         optimal_x = None
         num_restarts = 50
+        y_max = self.get_best_trials()[0].score
```
omalleyt12: The ```score``` is sometimes minimized and sometimes maximized. Also we do score normalization before fitting in the GPR

Instead this should use the vectorized trials, like:

```python
y_max = np.max(y)
```

kerastuner/tuners/bayesian.py
```diff
@@ -107,21 +112,19 @@ def _populate_space(self, trial_id):
...
         bounds = self._get_hp_bounds()
         x_seeds = self._random_state.uniform(bounds[:, 0], bounds[:, 1],
                                              size=(num_restarts, bounds.shape[0]))
+        ac = utils.UtilityFunction(kind=self.acq, beta=self.beta).utility
```
omalleyt12: Prefer more descriptive name, like ```self.acquisition_function```. Also the logic to select the acquisition function should happen in ```__init__```

omalleyt12:
@themrzmaster Are you still interested in moving forward with this PR?

themrzmaster:
Yes, i will take a look this weekend

omalleyt12:
Closing due to inactivity, @themrzmaster please feel free to reopen if you'd still like to work on this

PR #163, opened 2019-11-11
Added types to hyperparameters.

gabrieldemarmiesse:
Since we only need compatibility for python 3.6+, we can type functions.

Specifically I would like to type signatures of the public API. We have multiple benefits for doing that:

1) Documentation: The types will be displayed in the docs, in the signatures, giving more (structured) informations to the reader.
2) Working environments: people working with IDEs and type checking tools will have checks performed in real time. This will improve productivity.
3) Later on, we can perform run-time checks when the function is called (when the overhead will be negligible). There are several projets which can help us there: [typeguard](https://github.com/agronholm/typeguard), [enforce](https://github.com/RussBaz/enforce), [pytypes](https://pypi.org/project/pytypes/) and others...

2 and 3 can greatly allow us to simplify the checks of user inputs, and reduce the number of github issue created because of a misuse of the API.

### DIFF
--- a/kerastuner/engine/hyperparameters.py
+++ b/kerastuner/engine/hyperparameters.py
@@ -21,6 +21,7 @@ import contextlib
 import math
 import numpy as np
 import random
+from typing import List, Union, Optional, Any
 
 from tensorflow import keras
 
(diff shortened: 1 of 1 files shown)


omalleyt12:
Thanks for the PR! 

I think we may need to support Python 2 for internal use cases, I'll check on this and get back. If not, I agree adding this info would be very useful

gabrieldemarmiesse:
If we still need to support python 2 for Google internal uses cases, when can we expect to drop support for python 2? This is not the first time I have to revert some commits because of this and it starts to get a bit frustrating, especially since the there in so Python 2.7 build on travis.

See https://github.com/keras-team/keras-tuner/pull/57#discussion_r316321444 and https://github.com/keras-team/keras-tuner/pull/63 .

We need some clarity on the subject, and we need a strict rule written somewhere otherwise we are going to waste the time of people making changes which are going to get refused.

gabrieldemarmiesse:
Closing this since we need to support python 2.7 #179

omalleyt12:
Ah yeah, turns out we have to support Py2 for awhile after all lol

PR #176, opened 2019-12-04
Fix for multi-execution Tuner: ensure callbacks are run each execution

omalleyt12:
Previously the callbacks were only running for the first execution because they were popped from the `fit_kwargs` each execution


TreeKat71:
I think if it is fixed like this, then line 86 in the original is redundant
`fit_kwargs = copy.copy(fit_kwargs)`

PR #178, opened 2019-12-09
Allow proto of only values to be passed, create Fixed parameters

omalleyt12:
This change allows TFX to pass in simple values and create a HyperParameters object. Values are mapped only `Fixed` space objects.

This allows TFX to integrate Keras Tuner into its future tuner and trainer components, and this way all Python logic still ties together search space and values


omalleyt12:
Thanks for the review!

PR #186, opened 2019-12-15
Using docker to run the ci.

gabrieldemarmiesse:
It provides easy steps to reproduce whatever happens on travis ci locally.

It will help newcomers to run the test as well as ensure we can debug locally whatever happens in the CI.

The Dockerfile has been optimized for using the cache as much as possible. Locally, we'll benefit from it. Unfortunately, in travis-ci, the cache isn't saved from run to run.  This might be solved in the future by using `docker buildx`, but for the moment, the build times are acceptable.

For example, if you want to run the integration tests locally, either do it yourself by installing python, pytest, etc... or if you have docker, just do:

```bash
docker build -f tests/Dockerfile -t test_image --build-arg PY_VERSION=3.6 .
docker run -t test_image py.test tests/integration_tests
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ install:
   # code below is taken from https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/use-conda-with-travis-ci.html
   # We do this conditionally because it saves us some downloading if the
   # version is the same.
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Yes, it will. I pass the environment variable containing the python version to docker at build time. You can check it's the case on travis.

PR #189, opened 2019-12-17
Added types, compatible with py2.

gabrieldemarmiesse:
This is a follow-up on https://github.com/keras-team/keras-tuner/pull/163

I added types, but in a way compatible with python 2.7. See https://mypy.readthedocs.io/en/latest/python2.html for details on typing python 2 code.

### DIFF
--- a/kerastuner/engine/hyperparameters.py
+++ b/kerastuner/engine/hyperparameters.py
@@ -22,6 +22,8 @@ import math
 import numpy as np
 import random
 import six
+from typing import List, Optional, Any
+from pytypes import typechecked
 
 from tensorflow import keras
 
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Unrelated failure. Please merge #186 first.

omalleyt12:
Closed / reopened to retrigger tests

omalleyt12:
@gabrieldemarmiesse Thanks for the PR! Should we also include `typing` in dependencies for keras-tuner, or is this built-in?

gabrieldemarmiesse:
Good question. I need to check.

gabrieldemarmiesse:
Ok, so the bottom line is that typing is build-in in python 3.5+ and that it's not build in python 2.7. I need to add it to the setup.py.

gabrieldemarmiesse:
As said in the project description https://pypi.org/project/typing/: 

>  in Python 3.5 and later, the typing module lives in the stdlib, and installing this package has NO EFFECT. 

So the typing package is going to be installed in both python 2 and 3, but in python 3 it's going to have no effect. This is what we want.

haifeng-jin:
no longer supporting py2

PR #196, opened 2019-12-23
Class variables in HyperXception and HyperResnet changed | classes ch…

sriharshakonuru:
- Class variables in HyperXception and HyperResnet changed from classes to num_classes

- Difference in documentation and code resolved


omalleyt12:
@sriharshakonuru Thanks for the PR!

I think instead we should change the documentation. Using `classes` was a conscious decision to be inline with what Keras Applications does

PR #210, opened 2020-01-10
Set seed before each test.

gabrieldemarmiesse:
We had some issues with the CI about tests failing randomly. This PR fixes the seed before each test, allowing us to avoid random failure and to avoid contributors wondering what happened to the CI.


haifeng-jin (approved):
Thanks for the PR! LGTM.
Should be used for #482 .

PR #214, opened 2020-01-11
Download tf-cpu instead of tf-gpu

gabrieldemarmiesse:
tensorflow gpu is ~400MB and tensorflow cpu is ~100MB. This should speed up the install in the CI.


omalleyt12:
Thanks for the PR!

PR #215, opened 2020-01-11
Load mnist before starting the tests.

gabrieldemarmiesse:
We have some issues with downloading mnist during the tests. I propose here to make it a separate step so that it becomes more visible if the error comes from here in the CI. I added the retry on travis in case the connection breaks.


omalleyt12:
@gabrieldemarmiesse Thanks for the PR! Apologies, just saw this. I updated the integration test to use dummy data, since the data download was flaking so often

gabrieldemarmiesse:
No worries!

PR #216, opened 2020-01-13
Add Colabs for examples 

mrm8488:
I've created Google Colaboratory Notebooks for the two examples in examples folder (helloworld and CIFAR10)


haifeng-jin:
Thank you for the PR!
Unfortunately, we plan to rely on an autogenerate mechanism to generate notebooks from python files.
So we will not keep the manually wrote ipynb files in the repo.

PR #225, opened 2020-01-31
Update bayesian.py

ben-arnao:
Match the GPR Kernel loaded to the one used in initialization

### DIFF
--- a/kerastuner/tuners/bayesian.py
+++ b/kerastuner/tuners/bayesian.py
@@ -106,29 +106,42 @@ class BayesianOptimizationOracle(oracle_module.Oracle):
             values = self._random_trial()
             return {'status': trial_lib.TrialStatus.RUNNING,
                     'values': values}
-
+        
+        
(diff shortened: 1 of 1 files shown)


omalleyt12:
Going to close and reopen to retrigger the CI, the test failures look like flakes

ben-arnao:
It might be cleaner to just have a function to construct the gpr and put it in both places so there are no discrepancies

```
    def get_gpr(self):
        return gaussian_process.GaussianProcessRegressor(
            kernel=gaussian_process.kernels.Matern(nu=2.5),
            n_restarts_optimizer=20,
            normalize_y=True,
            alpha=self.alpha,
            random_state=self._seed_state)
```

omalleyt12:
> It might be cleaner to just have a function to construct the gpr and put it in both places so there are no discrepancies

@ben-arnao Apologies for the delayed response. Yes I agree that would be cleaner. Can you make that change and ping this thread?

Issue #224, opened 2020-01-30
GPR for Bayesian Opt initializes with Matern kernel but loads a different one?


ben-arnao:
Looks like the Bayesian Oracle is initialized with the following kernel for GPR

`kernel=gaussian_process.kernels.Matern(nu=2.5)`

But then if we load back up the Oracle the a different kernel is loaded

`kernel=gaussian_process.kernels.ConstantKernel(1.0)`

Is there a reason for this?

omalleyt12:
@yogi-dmt Thanks for the issue!

This looks like an oversight from when we changed kernels. Would you be able to submit a PR that changes this to use the Matern kernel when loading?

omalleyt12:
Fixed now

PR #229, opened 2020-02-03
update to fix issue #224

themrzmaster:
Add matern kernel to gp


omalleyt12:
Thanks for the PR! Sorry just saw this PR, looks like this is fixed now:

https://github.com/keras-team/keras-tuner/pull/243

PR #243, opened 2020-03-06
Make GPR consistent among load and create, put in one wrapper function

ben-arnao:
Put the GPR initialization code in one function and then add the wrapper to BayesianOptimizationOracle init and load as they should both be creating and loading the same base GPR instance. #224

### DIFF
--- a/kerastuner/tuners/bayesian.py
+++ b/kerastuner/tuners/bayesian.py
@@ -76,9 +76,9 @@ class BayesianOptimizationOracle(oracle_module.Oracle):
         self._tried_so_far = set()
         self._max_collisions = 20
         self._random_state = np.random.RandomState(self.seed)
-        self.gpr = self.make_gpr()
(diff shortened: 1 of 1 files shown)


kerastuner/tuners/bayesian.py
```diff
@@ -76,7 +76,10 @@ def __init__(self,
         self._tried_so_far = set()
         self._max_collisions = 20
         self._random_state = np.random.RandomState(self.seed)
-        self.gpr = gaussian_process.GaussianProcessRegressor(
+        self.gpr = self.make_gpr()
+        
+    def make_gpr(self):
```
omalleyt12: Let's make this private: `def _make_gpr`

ben-arnao: Should be good now.

omalleyt12:
@ben-arnao Thanks! Failures are unrelated, merging now

PR #246, opened 2020-03-07
Change init points

ben-arnao:
Let user define the dimensional factor for init points, also expose init points as a property (ie. if user wants to log when the optimization is using random trials and when it switches over to bayesian search)

### DIFF
--- a/kerastuner/tuners/bayesian.py
+++ b/kerastuner/tuners/bayesian.py
@@ -55,7 +55,8 @@ class BayesianOptimizationOracle(oracle_module.Oracle):
     def __init__(self,
                  objective,
                  max_trials,
-                 num_initial_points=None,
+                 init_points=None,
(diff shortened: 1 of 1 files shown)


ben-arnao:
@omalleyt12 Thanks for the feedback. I do see your point about cognitive overhead. My main thinking here was more so that users didn't have to go too much into the code if they wanted these things.

I do believe I ran into a scenario where if beta was too low then I would have to increase the initial points for BO to get going or else it would just get stuck on the first point for some reason. I'm not 100% sure though, so i will try to reproduce to verify.

To your point it can be hard for non BO users (and even BO users) to know how many init points to use, I think a lot will be relying on the default of 3x dimensions. Might be worth it to look into whether or not this is the most ideal. For an expensive target functions, extra or too little init points can be costly in the overall efficiency of the tuner.

Issue #254, opened 2020-03-16
Chief process does not gracefully exits 


sergiigladchuk:
Hi,
I am really new to the keras-tuner, but find it very useful. I was playing with "Distributed Tuning" tutorial from here (https://keras-team.github.io/keras-tuner/tutorials/distributed-tuning/) and chief process never exits when all the tuning options are tested. Is it by design?
Or am I missing something here? Should chief process exit when all the tuning options are exhausted?

I have checked `oracle_chief.py` there is just loop at the end:
```
server.start()
    while True:
        # The server does not block otherwise.
        time.sleep(10)
```
No exception catch to stop gRPC server and make it possible to gracefully exit. 

I ran same example from tutorial with reduced number of options for hyperparameters to try.

scfox:
I am seeing this same behavior.  After all tuners complete, they exit but chief process continues to run.

sergiigladchuk:
> I am seeing this same behavior. After all tuners complete, they exit but chief process continues to run.

Hopefully, maintainers will find time to help me to fix my PR and move the it forward. It is supposed to add at least possibility to send INT and other signals.

haifeng-jin:
Thank you for the issue.
This problem is fixed in the latest release of 1.0.2.
Let me know if it doesn't.

PR #255, opened 2020-03-17
KeyboardInterrupt for graceful exit

sergiigladchuk:
This is a small fix for #254

### DIFF
--- a/kerastuner/distribute/oracle_chief.py
+++ b/kerastuner/distribute/oracle_chief.py
@@ -16,6 +16,7 @@ from concurrent import futures
 import grpc
 import os
 import time
+import signal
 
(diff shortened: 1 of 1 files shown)


sergiigladchuk:
My first commit could not make it through tests/kerastuner/tuners/bayesian_test.py, which I did not touch. 
My later commits will not pass test_base_tuner_distribution, since I tried to use signal module, and it needs to execute signal handlers in the main Python thread. 
I hope either of the proposed solutions will be merged.

haifeng-jin:
It seems another fix is already in.
So I am closing this PR.
Thanks again for the PR!
Let me know if the master branch still doesn't solve the issue.

PR #269, opened 2020-04-02
Pass down tuner_id from Tuner class constructor

ucdmkt:
There was an unused, undocumented named constructor `tuner_id` arg to Tuner class. I believe this was meant to override tuner_id, if given.


ucdmkt:
\cc @yinghsienwu

Issue #290, opened 2020-05-08
hello world fails:  "ValueError: Unknown parameter: num_layers"


carchrae:
i was experimenting with the hello world example, and sadly it throws an error on the last example

https://github.com/keras-team/keras-tuner/blob/d2d69cba21a0b482a85ce2a38893e2322e139c01/examples/helloworld.py#L189

```
INFO:tensorflow:Reloading Oracle from existing project ./untitled_project/oracle.json
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/kerastuner/engine/hypermodel.py", line 105, in build
    model = self.hypermodel.build(hp)
  File "<ipython-input-4-a999c064421f>", line 185, in build_model
    for i in range(hp.get('num_layers')):
  File "/usr/local/lib/python3.6/dist-packages/kerastuner/engine/hyperparameters.py", line 617, in get
    'Unknown parameter: {}'.format(full_name))
ValueError: Unknown parameter: num_layers
[Warning] Invalid model 0/5
```

to ensure it wasn't something funny about my local environment, i tried it on colab too (same error).

anyway, my investigating leads me to believe this is due to the caching of previous trials (a clever but foot-shooty feature).  the issue is that it loads a previous trial (where `num_layers` was not defined as a parameter) and so it blows up because `hp` has not such property in it. 

the fix is pretty simple, just blow away the old saved trials with `overwrite=True`

```
tuner = RandomSearch(
    build_model,
    max_trials=5,
    hyperparameters=hp,
    allow_new_entries=False,
    objective='val_accuracy',
    overwrite=True)
```

i'll send a PR in a sec

carchrae:
i do wonder though if it would not be better to have each example save to a unique directory.  eg, not use ``directory='test_dir'` and also set `project_name` to something unique.  i will make another PR for that.

carchrae:
:point_up: i think #292 is a nicer solution.  it also cleans up the code so it puts all the test results in `test_dir`

PR #291, opened 2020-05-08
overwrite model since new parameter added (num_layers)

carchrae:
fix for #290


carchrae:
i'm going to close this in favour of #292 which i think is a better fix

PR #294, opened 2020-05-08
Skip known failure

### LINKED ISSUE #293: fix known test failures
trying to be a good oss citizen, i spent time to try and understand why my PR was causing a test failure ( `test_override_compile fails` ) .  it turns out that it isn't my code change at all and this test has been failing for a while.

to avoid wasting people's time, i suggest skipping this test.  or fixing it.  fixing it is not feasible for me right this moment, so i submit a pr to put your build back to passing.

update: i found a second test that fails intermittently (due to randomness) and so i have also submitted a PR to make these failures less likely.  (see below)

carchrae:
resolves #293


haifeng-jin:
All the failure in the tests has already been fixed.
I am closing the PR.
Thank you again for your contribution!

PR #303, opened 2020-05-27
add alpha and beta to Oracle object creation

sim-san:
The parameters alpha and beta are now passed to the Orcale Object so that these values can be changed.


sim-san:
Closed by https://github.com/keras-team/keras-tuner/pull/349

PR #304, opened 2020-05-28
hyperbrand max iterations check typo fix

### LINKED ISSUE #256: Distributed hyperband tuner runs continuously
Hello,

I ran the distributed tuner tutorial on google cloud virtual machines with a shared filesystem. One chief instance with no GPU, multiple worker instances with a single GPU. When I run the tutorial with a single worker and for one complete hyperband iteration, it runs successfully. When I run it with two workers, it looks like the workers share the workload initially, but then one exits, and the other worker keeps going and starts running through earlier brackets of the hyperband algorithm.

Has anybody seen anything like this? Is there something I am missing?

Tomer23:
resolves #256

In a distributed setup if you have more worker - or start the same worker twice - the first exists as it reaches the max iterations limit but the second goes over the max iterations bound and continuously creates new and new brackets in an endless loop. It can be solved with that:

In tuners/hyperbrand.py 174
`
        if (self._current_bracket == 0 and
                self._current_iteration + 1 >= self.hyperband_iterations):`

### DIFF
--- a/kerastuner/tuners/hyperband.py
+++ b/kerastuner/tuners/hyperband.py
@@ -197,6 +197,8 @@ class HyperbandOracle(oracle_module.Oracle):
         if self._current_bracket < 0:
             self._current_bracket = self._get_num_brackets() - 1
             self._current_iteration += 1
+            if self._current_iteration > self.hyperband_iterations:
+                self._current_bracket = 0
 
     def _remove_completed_brackets(self):
         # Filter out completed brackets.
--- a/kerastuner/tuners/hyperband.py
+++ b/kerastuner/tuners/hyperband.py
@@ -172,7 +172,7 @@ class HyperbandOracle(oracle_module.Oracle):
 
         # Max sweeps has been reached, no more brackets should be created.
         if (self._current_bracket == 0 and
-                self._current_iteration + 1 == self.hyperband_iterations):
+                self._current_iteration + 1 >= self.hyperband_iterations):
             # Stop creating new brackets, but wait to complete other brackets.
             if self.ongoing_trials:
                 return {'status': 'IDLE'}


Tomer23:
I added a fix to stop workers to repeat the process for each worker (since the bracket number resets to start)

Tomer23:
I rebased it because the tests seemed to be faulty.

sim-san:
@haifeng-jin @omalleyt12 @yixingfu  Can somebody review an merge this PR ?

kerastuner/tuners/hyperband.py
```diff
@@ -172,7 +172,7 @@ def _populate_space(self, trial_id):
 
         # Max sweeps has been reached, no more brackets should be created.
         if (self._current_bracket == 0 and
```
haifeng-jin, line 174: Would it be better to change this check to "<=" instead of setting it to 0 when it its negative?

haifeng-jin:
@Tomer23 @sim-san sorry for the late reply. Please see my comments.

sim-san:
> 
> 
> @Tomer23 @sim-san sorry for the late reply. Please see my comments.

@haifeng-jin Thanks for reviewing
@Tomer23 I think you can answer the code question better

kerastuner/tuners/hyperband.py
Tomer23: I don't think it would work, the problem is not that it can be negative. 
In 180 it would call to increment it, but the increment (195-199) will not let it go to negative, it would restart before that. A negative _current_bracket number can never reach line 174 (bc 197).
My solution would stop the restarting.

If you want it that way you suggested:
```suggestion
        if (self._current_bracket <= 0 and
```
you also need in line 197
```
        if self._current_bracket < 0 and self._current_iteration <= self.hyperband_iterations:
```

You need to stop the restarting either way.

PR #312, opened 2020-06-01
Control trial output verbosity

Aviously:
Allows individual trial outputs to be suppressed during a search by specifying `verbose=0`.

### DIFF
--- a/kerastuner/engine/base_tuner.py
+++ b/kerastuner/engine/base_tuner.py
@@ -115,7 +115,8 @@ class BaseTuner(stateful.Stateful):
             *fit_kwargs: Keyword arguments that should be passed to
               `run_trial`, for example the training and validation data.
         """
-        verbose = fit_kwargs.get('verbose', 1)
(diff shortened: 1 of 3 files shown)


kerastuner/engine/base_tuner.py
```diff
@@ -126,9 +127,9 @@ def search(self, *fit_args, **fit_kwargs):
                 # Oracle is calculating, resend request.
                 continue
 
-            self.on_trial_begin(trial)
+            self.on_trial_begin(trial, verbose)
```
omalleyt12, line 130: I think rather than adding arguments to `on_trial_begin` and `on_trial_end`, it might make more sense to do `self._display.verbose = verbose` and then have the display contain the logic for what it should/shouldn't print

That way the logging logic is more self-contained in the display object. WDYT?

Aviously: Sounds good! I just made an update with these changes.

omalleyt12:
Failures are unrelated, merging now

PR #314, opened 2020-06-05
Fix #284. Parent condition should not be checked.

### LINKED ISSUE #284: `HyperParameter` not defined.
I am building model generator that uses 3 levels of parameters

1. Decides how many **additional LSTM layers** to be added:

```
hp.Int(name='Additional_LSTM_n', min_value=0, max_value=2, step=1)
```

0 = zero layers, 1= one layer, 2 = 2 layers

2. Decides if the LSTM level should have a **Dropout** layer added. **Depends on previous parameter** value. **It  should either be 1 or 2.**
```
hp.Boolean(name=f'Droupout_LSTM_1', parent_name='Additional_LSTM_n', parent_values=[1, 2])
```
3. Decides what kind of **rate** should be used for the dropout layer. **Depends on the previous parameter.** Should only be used if previous parent value is **True**.

```
hp.Choice(name='Dropout_1_rate', values=[0.1, 0.2], default=0.2, parent_name=f'Droupout_LSTM_1', parent_values=True)
```

Upon execution I get following error:

> ValueError: `HyperParameter` named: Droupout_LSTM_1 not defined.

the code to reproduce the issue is:

```
    hp = hp_module.HyperParameters()

    hp.Int(name='Additional_LSTM_n', min_value=0, max_value=2, step=1)
    hp.Boolean(name=f'Droupout_LSTM_1', parent_name='Additional_LSTM_n', parent_values=[1, 2])
    hp.Choice(name='Dropout_1_rate', values=[0.1, 0.2, 0.3, 0.4, 0.5], default=0.2, parent_name=f'Droupout_LSTM_1', parent_values=True)

    print(hp.values)
```

What am I doing wrong? Are the 3-level nested parameters not supported? 

**It is important that step 3. is not run if the step 2. is False. I don't want to end up in a situation when the tuner generates the value for 3. but it is never used because 2. is false.. Otherwise some executions will be duplicated, wasting the GPU usage, time, and not producing any "true" new models.**

yixingfu:
This error is because when registering new parameter, checking the existence of the parent parameter requires the condition of current scope to match with the condition of the parent's, which is always false.

### DIFF
--- a/kerastuner/engine/hyperparameters.py
+++ b/kerastuner/engine/hyperparameters.py
@@ -566,7 +566,7 @@ class HyperParameters(object):
               HyperParameters under this scope should be considered active.
         """
         parent_name = self._get_name(parent_name)  # Add name_scopes.
-        if not self._exists(parent_name):
+        if not self._exists(parent_name, conditions='any'):
             raise ValueError(
                 '`HyperParameter` named: ' + parent_name + ' '
                 'not defined.')
@@ -602,6 +602,9 @@ class HyperParameters(object):
 
     def _exists(self, name, conditions=None):
         """Checks for a `HyperParameter` with the same name and conditions."""
+        if conditions == 'any':
+            return name in self._hps
+
         if conditions is None:
             conditions = self._conditions
 
--- a/tests/kerastuner/engine/hyperparameters_test.py
+++ b/tests/kerastuner/engine/hyperparameters_test.py
@@ -499,3 +499,11 @@ def test_prob_one_choice():
 
     value = hp_module.cumulative_prob_to_value(0, hp)
     assert value == 0
+
+
+def test_nested_parent():
+    hp = hp_module.HyperParameters()
+    hp.Choice('a', [1, 2, 3], default=1)
+    hp.Choice('b', [1, 2, 3], default=2, parent_name='a', parent_values=1)
+    hp.Choice('c', [1, 2, 3], default=3, parent_name='b', parent_values=1)
+    assert hp.values == {'a': 1, 'b': 2}


qlzh727:
@omalleyt12, what's the process of reviewing PR for this repo? Is there any active member monitoring the incoming PR and triage them?

omalleyt12:
@qlzh727 Yep I'll take a look now, thanks!

omalleyt12:
@yixingfu Thanks for the PR!

Could you provide some more context on what issue this is solving?

yixingfu:
I tried to solve #284 , where checking the existence of parent with `_exists()` marks the parent that is in itself conditional falsely conclude that the parent does not exist.
Say, we have created 'a', and 'b' conditioned on 'a', and try to create 'c' that conditions on 'b'. Then when creating 'c', the scope does not have the condition of 'b' in `_conditions` and hence reject the creation of 'c'. I think the simplest fix is to skip this condition check when determine whether 'b' exist.

yixingfu:
So actually the real problem here is that specifying parent using keyword does not carry any information about the conditions on the parent. 

After some more thinking, it looks like discerning variables with the same name but different conditions is important. Hence it may be better to not make the change I previously PRed; instead, it should just be made clear that when trying to create parameters with nested parents, the scope should be used explicitly instead of simply setting `parent_name` and `parent_values` keywords.

Otherwise we may need to add a keyword like 'parent_conditions' to pass down this information?

yixingfu:
@omalleyt12 
I think I have question on the purpose of keeping track of and distinguishing hyperparameters with the same name but different conditions.

To me the purpose of distinguishing such hyperparemeters is to allow situations like following:
```
hp.Int('a', 1, 2)
with hp.conditional_scope('a', 1):
    hp.Int('b', 2, 3)
    with hp.conditional_scope('b', 3):
        hp.Int('c', 4, 5)
with hp.conditional_scope('a', 2):
    hp.Int('b', 2, 3)
    with hp.conditional_scope('b', 2):
        hp.Int('c', 4, 5)
```
However, the two different `b`s cannot be accessed using the same footing, and has to be treated separately (i.e. I cannot simply use a `b=hp.Int('b')` or something to use all situations of `b`). So, wouldn't it be better to disallow any conflicting definition of hyperparameter under same name?

If hyperparameter has one-to-one correspondence with their names, parents only need to be checked with their names and no other condition, and keywords `parent_name` and `parent_values` carry all information needed. Then, nested parents can be easily supported by allowing more than one parent name-value pairs .

ssavale-entrata:
To support nested parents, Is it good to make changes in function _exists & conditional_scope like below way? 

class HyperParameters(object):

    def _exists(self, name, conditions=None):
        """Checks for a `HyperParameter` with the same name and conditions."""
        if conditions is None:
            conditions = self._conditions

        if name in self._hps:
            hps = self._hps[name]
            for hp in hps:
                if not hp.conditions:
                    return True
                else:
                    for hp_condition in hp.conditions:
                        for condition in conditions:
                            if hp_condition == condition:
                                return True
        return False
		
	@contextlib.contextmanager
    def conditional_scope(self, parent_name, parent_values):
        """Opens a scope to create conditional HyperParameters.

        All HyperParameters created under this scope will only be active
        when the parent HyperParameter specified by `parent_name` is
        equal to one of the values passed in `parent_values`.

        When the condition is not met, creating a HyperParameter under
        this scope will register the HyperParameter, but will return
        `None` rather than a concrete value.

        Note that any Python code under this scope will execute
        regardless of whether the condition is met.

        # Arguments:
            parent_name: The name of the HyperParameter to condition on.
            parent_values: Values of the parent HyperParameter for which
              HyperParameters under this scope should be considered active.
        """
        parent_name = self._get_name(parent_name)  # Add name_scopes.
        if not self._exists(parent_name):
            raise ValueError(
                '`HyperParameter` named: ' + parent_name + ' '
                'not defined.')

        self._conditions.append(conditions_mod.Parent(parent_name, parent_values))
        yield

        """
        try:
            yield
        finally:
            self._conditions.pop()
        """

ssavale-entrata (commented):
I have implemented this change at my local environment and everything is working perfectly, I am eagerly waiting for this change, please release it as soon as possible.

haifeng-jin:
@yixingfu Would you like to sign the

haifeng-jin:
@ssavale-entrata I have commented on #284. You may take a look. Thanks.

PR #318, opened 2020-06-11
do not delete best epoch checkpoint

haifeng-jin:


omalleyt12:
Thanks for the PR! Merging now

PR #325, opened 2020-06-16
stop server after finish

yixingfu:
When running distributed search, the chief_oracle server runs indefinitely after search finishes. This will hold the job running until someone manually tears it down. 

What I do is I add an attribute in `OracleServicer` that is set to True when `chief_oracle` is giving out `trial` with stopped status. Then in the loop that block the server, once the stopping is triggered it will wait for all ongoing trials to end before stopping the server and escape from the loop.

### DIFF
--- a/kerastuner/distribute/oracle_chief.py
+++ b/kerastuner/distribute/oracle_chief.py
@@ -85,7 +85,6 @@ def start_server(oracle):
 
         if oracle_servicer.stop_triggered:
             while oracle.ongoing_trials:
-                print(f'Stop is triggered. Remaining open trials: {oracle.ongoing_trials}.')
                 time.sleep(10)
 
             print('Exiting in 10s.')
--- a/kerastuner/distribute/oracle_chief.py
+++ b/kerastuner/distribute/oracle_chief.py
@@ -87,6 +87,7 @@ def start_server(oracle):
             while oracle.ongoing_trials:
                 time.sleep(10)
 
-            print('Exiting in 10s.')
+            print('Oracle server on chief is exiting in 10s.'
+                  'The chief will go on with post-search code.')
             server.stop(10)
             break
--- a/kerastuner/distribute/oracle_chief.py
+++ b/kerastuner/distribute/oracle_chief.py
@@ -27,6 +27,7 @@ class OracleServicer(service_pb2_grpc.OracleServicer):
 
     def __init__(self, oracle):
         self.oracle = oracle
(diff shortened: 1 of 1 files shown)


kerastuner/distribute/oracle_chief.py
```diff
@@ -71,10 +74,20 @@ def start_server(oracle):
...
+        time.sleep(30)
+
+        if oracle_servicer.stop_triggered:
+            while oracle.ongoing_trials:
+                print(f'Stop is triggered. Remaining open trials: {oracle.ongoing_trials}.')
+                time.sleep(10)
+
+            print('Exiting in 10s.')
+            server.stop(10)
+            break
```
omalleyt12, line 93: Should we raise an error after stopping the Oracle server?

The reason is, we expect users to write their code like:

```
tuner = Tuner(...). # Runs an Oracle server if KERASTUNER_TUNER_ID='chief' environment variable is set
tuner.search(...)
```

Because of this, if the Oracle server stops, the tuner will try and run `tuner.search` but it won't work

Maybe we should raise a `StopIterationError` to prevent this? It will have the unfortunate side effect of having the chief oracle job marked as failure though, so i'm not sure. Wdyt?

yixingfu: When the tuner tries tuner.search it, if `overwritten` is not set it will finish searching right away.  That is what I see when I tried it.  
Raising an error may indeed be safer, though?

kerastuner/distribute/oracle_chief.py
```diff
@@ -71,10 +74,20 @@ def start_server(oracle):
...
     server.add_insecure_port('{}:{}'.format(ip_addr, port))
     server.start()
     while True:
         # The server does not block otherwise.
-        time.sleep(10)
+        time.sleep(30)
+
+        if oracle_servicer.stop_triggered:
+            while oracle.ongoing_trials:
+                print(f'Stop is triggered. Remaining open trials: {oracle.ongoing_trials}.')
```
omalleyt12, line 88: Maybe we should remove this to avoid printing a lot of lines

pooria121:
is this been merged ? and if we install keras-tuner through pip would get this update?

yixingfu:
> is this been merged ? and if we install keras-tuner through pip would get this update?

It is merged but pip does not get the update immediately. You can try `pip install git+https://github.com/keras-team/keras-tuner@master`.

Issue #308, opened 2020-05-29
Keras Tuner stopped at the end of first epoch when train on TPU


qlzh727:
This is discovered during test of https://www.kaggle.com/kivlichangoogle/jigsaw-multilingual-getting-started. Note that this issue only happens when TPU is enabled for the notebook.

When Keras tuner is added, the search process will fail at the start of send epoch, when saving/loading checkpoint. I think the root cause is in TF and TPU, but a temp walk around in keras tuner will be nice, eg disable checkpointing if possible.

The error stack is like below:

```
---------------------------------------------------------------------------
UnimplementedError                        Traceback (most recent call last)
<ipython-input-14-a9734e750f48> in <module>
      6              verbose=1,
      7              validation_data=nonenglish_val_datasets['Combined'],
----> 8              validation_steps=100)

/opt/conda/lib/python3.7/site-packages/kerastuner/engine/base_tuner.py in search(self, *fit_args, **fit_kwargs)
    128 
    129             self.on_trial_begin(trial)
--> 130             self.run_trial(trial, *fit_args, **fit_kwargs)
    131             self.on_trial_end(trial)
    132         self.on_search_end()

/opt/conda/lib/python3.7/site-packages/kerastuner/engine/multi_execution_tuner.py in run_trial(self, trial, *fit_args, **fit_kwargs)
     94 
     95             model = self.hypermodel.build(trial.hyperparameters)
---> 96             history = model.fit(*fit_args, **copied_fit_kwargs)
     97             for metric, epoch_values in history.history.items():
     98                 if self.oracle.objective.direction == 'min':

/opt/conda/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/training.py in fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_freq, max_queue_size, workers, use_multiprocessing, **kwargs)
    817         max_queue_size=max_queue_size,
    818         workers=workers,
--> 819         use_multiprocessing=use_multiprocessing)
    820 
    821   def evaluate(self,

[...]

PR #328, opened 2020-06-19
Allow save_freq keyword for search

yixingfu:
Pass down `save_freq` keyword from `search()` to the checkpointing callback in `run_trial`. This allows user to request less frequent (or even completely disable) saving epochs.

In case the best epoch (or even any epoch) is saved, at `load_model` any exception in loading weights is caught and give a warning saying the model need to retrained. This may be helpful in general, when checkpoint become unavailable for any reason. Then user will be able to still get the model with best hyperparameter, and retrain the model.

Allowing no-checkpoint and failed weight-loading provide work around for #308, and potentially also other issues caused by filesystem.

### DIFF
--- a/kerastuner/engine/multi_execution_tuner.py
+++ b/kerastuner/engine/multi_execution_tuner.py
@@ -72,13 +72,18 @@ class MultiExecutionTuner(tuner_module.Tuner):
         pass
 
     def run_trial(self, trial, *fit_args, **fit_kwargs):
+        save_freq = 'epoch'
+        if 'save_freq' in fit_kwargs:
+            save_freq = int(fit_kwargs.pop('save_freq'))
+
         model_checkpoint = keras.callbacks.ModelCheckpoint(
             filepath=self._get_checkpoint_fname(
                 trial.trial_id, self._reported_step),
             monitor=self.oracle.objective.name,
             mode=self.oracle.objective.direction,
             save_best_only=True,
-            save_weights_only=True)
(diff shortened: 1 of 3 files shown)


yixingfu:
@omalleyt12

yixingfu:
Clicked close accidentally.

kerastuner/engine/multi_execution_tuner.py
```diff
@@ -72,13 +72,18 @@ def on_epoch_end(self, trial, model, epoch, logs=None):
         pass
 
     def run_trial(self, trial, *fit_args, **fit_kwargs):
+        save_freq = 'epoch'
+        if 'save_freq' in fit_kwargs:
+            save_freq = int(fit_kwargs.pop('save_freq'))
+
         model_checkpoint = keras.callbacks.ModelCheckpoint(
```
omalleyt12, line 79: Hm maybe instead, we could check to see if the user passed their own `ModelCheckpoint` object? If they did, instead of creating our own, we could patch the save directory on that object so that it saves to a different directory each time, and use that?

We do something similar for TensorBoard here: https://github.com/keras-team/keras-tuner/blob/master/kerastuner/engine/tuner.py#L272

yixingfu: I see. Asking user to pass `ModelCheckpoint` object will be clearer and more consistent than allowing a specific keyword. Is that what you mean?

Issue #326, opened 2020-06-16
Behavior of bayesian tuner when initial points have same score


yixingfu:
This test case `test_bayesian_oracle_with_zero_y` (https://github.com/keras-team/keras-tuner/blob/master/tests/kerastuner/tuners/bayesian_test.py#L51) fails not necessarily with y=0; but whenever the randomly chosen initial points have the same score. Having identical `y` makes normalization of `y` vector singular and produce NaN. 

This should be a simple fix.  But what behavior should we expect when such situation happens? I can think of a few possibilities:
1. Do one more random trial until the Gaussian regression works;
2. Not normalizing y, either (a) for all cases or (b) whenever the singular situation happens;
3. Perturb score slightly for the initial points to avoid identical scores.

ben-arnao:
I mean the real question is why do all of your initialization points have the same score. I think that is usually indicative of something being wrong with the target function. However if you do have a reason to believe this is normal and fine, we can make a slight adjust to continue generating random points until all scores are not equal.

In `kerastuner/tuners/bayesian.py`

```
if len(completed_trials) < num_initial_points or len(set([trial.score for trial in self.trials])) <= 1:
    return self._random_populate_space()
```

yixingfu:
@ben-arnao Yes I agree with you, and that is why I think this test case is highly unnatural.
I just thought removing a test case because it fails sounds like a bad practice, so I tried to come up a more consistent way to get around the test case. As far as I see, what you suggest is similar to approach (1) I mentioned.

PR #348, opened 2020-07-07
fix test case bayesian with zero y

yixingfu:
The test case bayesian with zero y is highly unnatural. 
Here I fix it by using random parameters when scores for the initial points are exactly the same. 
Because it is unnatural, having repeating score likely hint something wrong. Hence warning message is shown. 

#326


yixingfu:
Redundant with a part of #361 . Closing this.

PR #353, opened 2020-07-11
Use default hp value only when not populated

haifeng-jin:
AutoKeras need to know the space in advance and populate even some unreached hp values.
This PR let the hypermodel use those populated values if available instead of the default values.

### DIFF
--- a/kerastuner/engine/hyperparameters.py
+++ b/kerastuner/engine/hyperparameters.py
@@ -636,8 +636,10 @@ class HyperParameters(object):
         value = hp.default
         # Only add active values to `self.values`.
         if self._conditions_are_active(hp.conditions):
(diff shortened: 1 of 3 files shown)


yixingfu:
@haifeng-jin This change breaks unit test `tests/kerastuner/engine/hyperparameters_test.py::test_merge`.
When merging, there is a preexisting value which is not default value. How can we reconcile this?

haifeng-jin:
@yixingfu Sorry for breaking the unit tests. I have a patch now. Running tests locally. I will pull request when it finishes.

haifeng-jin:
@yixingfu I created the PR here #361 .
It fixed all the broken tests.

Thanks.

Issue #356, opened 2020-07-15
Result summary of tuner includes trails with NaN score


nikvaessen:
Hi,

The tuner does not filter out trails with a NaN objective when using `Tuner#get_best_trails` or `Tuner#results_summary`.

See the attached colab as an example.

https://colab.research.google.com/drive/1FIbcELszorAEqKH_fef7U224-CKpyJx4?usp=sharing

spors:
I came across the same problem. A quick and dirty solution was to filter out the Nan scores:

```
def get_best_models(tuner, num_models=1):
        
    best_trials = tuner.oracle.get_best_trials(num_models)
    models = [tuner.load_model(trial) for trial in best_trials if not np.isnan(trial.score)]

    return models
```

PR #357, opened 2020-07-15
fix Oracle#get_best_trails returning trials with NaN score

nikvaessen:
See https://github.com/keras-team/keras-tuner/issues/356


haifeng-jin:
I think it would cause not returning the nan trials, which is not expected behavior.
For fixing the bug, it should return an nan trial without error.

PR #361, opened 2020-07-18
fix all broken tests

haifeng-jin:
Fixed all the failed tests.
Removed python2.7 test
Autodoc doesn't support python2.7 and there is a test using Autodoc.
Keras has different behaviors on metrics in python2 and python3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: python
 services:
   - docker
 python:
-  - "2.7"
   - "3.6"
 env:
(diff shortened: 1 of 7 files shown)


haifeng-jin:
One more issue here it seems the tests have some randomness. It has a chance to fail for some assertion like `assert result <= 0.9`. We may try to fix the seed in the future.

omalleyt12:
This is really great, thank you @haifeng-jin

PR #364, opened 2020-07-21
add application: HyperModel for augmentation

yixingfu:
This PR implements hyper model for augments.  

HyperAugment class collects the transforms to be used in each of the subclasses. Two HyperModels are implemented:

A) HyperFixedAugment: for each of the transforms assign a tuning knob.
B) RandAugment: https://arxiv.org/abs/1909.13719

These augments are only using preprocessing layers, and can be inserted as a part of other hypermodels. This usage is supported in efficientnet hypermodel (#363) and is crucial for using the hypermodel.

### DIFF
--- a/kerastuner/applications/augment.py
+++ b/kerastuner/applications/augment.py
@@ -101,7 +101,7 @@ class HyperFixedAugment(HyperAugment):
     """
 
     def build(self, hp):
-        model = keras.Sequential(name='FixedAugment')
+        model = keras.Sequential(name='fixed_augment')
 
         if self.input_tensor is not None:
             model.add(self.input_tensor)
@@ -186,6 +186,6 @@ class HyperRandAugment(HyperAugment):
                 # selection matches the transform index `i`
                 x = tf.where(tf.equal(i, selection), x_trans, x)
                 
-        model = keras.Model(inputs, x, name='RandAugment')
+        model = keras.Model(inputs, x, name='rand_augment')
         return model
 
--- a/kerastuner/applications/augment.py
+++ b/kerastuner/applications/augment.py
@@ -31,10 +31,10 @@ import random
 # Each function takes a factor (0 to 1) for the strength
 # of the transform.
 TRANSFORMS = {
-'translate_x': lambda x: preprocessing.RandomTranslation(x, 0),
-'translate_y': lambda y: preprocessing.RandomTranslation(0, y),
-'rotate': preprocessing.RandomRotation,
-'contrast': preprocessing.RandomContrast,
+    'translate_x': lambda x: preprocessing.RandomTranslation(x, 0),
+    'translate_y': lambda y: preprocessing.RandomTranslation(0, y),
+    'rotate': preprocessing.RandomRotation,
+    'contrast': preprocessing.RandomContrast,
 }
 
 class HyperAugment(hypermodel.HyperModel):
@@ -52,12 +52,12 @@ class HyperAugment(hypermodel.HyperModel):
             HyperModels. See `kerastuner.HyperModel`.
     """
     def __init__(self,
-        input_shape=None,
-        input_tensor=None,
-        rotate=True,
-        translate_x=True,
-        translate_y=True,
-        contrast=True):
+                 input_shape=None,
+                 input_tensor=None,
+                 rotate=True,
+                 translate_x=True,
+                 translate_y=True,
+                 contrast=True):
         self.transforms = []
         if rotate:
             self.transforms.append('rotate')
@@ -153,8 +153,8 @@ class HyperRandAugment(HyperAugment):
                  input_tensor=None,
                  **kwargs):
         super(HyperRandAugment, self).__init__(input_shape=input_shape,
-            input_tensor=input_tensor,
-            **kwargs)
+                                               input_tensor=input_tensor,
+                                               **kwargs)
         if input_shape is None and input_tensor is None:
             raise ValueError('You must specify either `input_shape` or '
                              '`intput_tensor` when using HyperRandAugment.')
@@ -171,12 +171,11 @@ class HyperRandAugment(HyperAugment):
         randaug_count = hp.Int('randaug_count', 0, 5, default=2)
 
         for _ in range(randaug_count):
-            # selection tensor determines for each sample, which operation
-            # is used.
+            # selection tensor determines operation for each sample.
             batch_size = tf.shape(x)[0]
             selection = tf.random.uniform([batch_size, 1, 1, 1],
-                maxval=len(self.transforms),
-                dtype='int32')
+                                          maxval=len(self.transforms),
+                                          dtype='int32')
 
             for (i, transform) in enumerate(self.transforms):
                 transform_layer = TRANSFORMS[transform](randaug_mag)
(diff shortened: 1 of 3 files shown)


qlzh727:
@omalleyt12 for review.

kerastuner/applications/augment.py
```diff
@@ -0,0 +1,191 @@
...
+from tensorflow.keras import layers
+from tensorflow.keras.layers.experimental import preprocessing
+
+import random
+
+# dict of functions that create layers for transforms.
+# Each function takes a factor (0 to 1) for the strength
+# of the transform.
+TRANSFORMS = {
+'translate_x': lambda x: preprocessing.RandomTranslation(x, 0),
```
qlzh727, line 34: Please add some indent for each item in the dict.

kerastuner/applications/augment.py
```diff
@@ -0,0 +1,191 @@
...
+            `layers.Input()`) to use as image input for the model.
+        rotate: Whether to include rotation in augmentation.
+        translate_x: Whether to include horizontal translation in augmentation.
+        translate_y: Whether to include vertical translation in augmentation.
+        contrast: Whether to include random contrast in augmentation.
+        **kwargs: Additional keyword arguments that apply to all
+            HyperModels. See `kerastuner.HyperModel`.
+    """
+    def __init__(self,
+        input_shape=None,
```
qlzh727, line 55: The indent for those params seems very broken.

kerastuner/applications/augment.py
```diff
@@ -0,0 +1,191 @@
...
+        if rotate:
+            self.transforms.append('rotate')
+        if translate_x:
+            self.transforms.append('translate_x')
+        if translate_y:
+            self.transforms.append('translate_y')
+        if contrast:
+            self.transforms.append('contrast')
+
+        self.input_shape = input_shape
```
qlzh727, line 71: Do u need to check whether only one of the input_shape nad input_tensor is provided?

kerastuner/applications/augment.py
```diff
@@ -0,0 +1,191 @@
...
+             pool of augmentations to select from.
+        translate_x: Whether to include horizontal translation in augmentation.
+        translate_y: Whether to include vertical translation in augmentation.
+        contrast: Whether to include random contrast in augmentation.
+        **kwargs: Additional keyword arguments that apply to all
+            HyperModels. See `kerastuner.HyperModel`.
+
+    # Search space:
+        'factor_{transform}' for each of the transforms chosen:
+            Float, default range [0.05, 1]. Controls the strength
```
qlzh727, line 99: Should we allow user to config this?

kerastuner/applications/augment.py
```diff
@@ -0,0 +1,191 @@
...
+            x = inputs
+
+        randaug_mag = hp.Float('randaug_mag', 0.05, 1.0, step=0.1, default=0.2)
+        randaug_count = hp.Int('randaug_count', 0, 5, default=2)
+
+        for _ in range(randaug_count):
+            # selection tensor determines for each sample, which operation
+            # is used.
+            batch_size = tf.shape(x)[0]
+            selection = tf.random.uniform([batch_size, 1, 1, 1],
```
qlzh727, line 177: I am bit confused about why this need to be a 4D tensor.

kerastuner/applications/augment.py
yixingfu: This can be controlled by overriding hyperparameters outside, for example
```
hp = HyperParameters()
hp.Float('factor_translate_x', 0, 0.1, step=0.01)
```
where hp will be used on constructing `tuner`.

It may also be useful to allow controlling the ranges using keyword arguments. Do you think it would be better to 1) add keywords for ranges corresponding to each transform; or 2) allow the keyword for each transform to be boolean, number or a tuple, which overrides the default config if set as number/tuple?

kerastuner/applications/augment.py
yixingfu: RandAugment selects a random operation on each sample in a batch (This is not very clearly and directly stated in the paper, but can be inferred from their [code](https://github.com/tensorflow/tpu/blob/master/models/official/efficientnet/autoaugment.py)), so this random number `selection` should be generated on the batch dimension.

kerastuner/applications/augment.py
yixingfu: `HyperFixedAugment` only uses sequential model so `input_shape` and `input_tensor` can be both missing. So the check is done only for `HyperRandAugment`.

kerastuner/applications/augment.py
qlzh727: Injecting items outside of augment model seems to be confusing, on the order hand, injecting a bunch of kwargs is also errorprone. @omalleyt12, do u have any recommendation here?

qlzh727:
I think this PR looks good to me, please wait for the review from @omalleyt12.

yixingfu:
@qlzh727 @omalleyt12 
I added functionality to allow configuring the search space. 

In order to config `HyperImageRandAugment` and `HyperImageFixedAugment` consistently (so that when we add more transform we only need to change `HyperImageAugment` instead of each of the subclasses ), I decide to allow a range of augment factor to be passed in for each keyword. When only one number is passed, it is understood as [0, x]. Then in general, when the range for the factor of certain transform is [a, b],  the factor will be given by a + (b-a) * `randaug_mag`.

kerastuner/applications/augment.py
```diff
@@ -0,0 +1,303 @@
...
+                                        default=factor_min)
+            if transform_factor == 0:
+                continue
+            transform_layer = TRANSFORMS[transform](transform_factor)
+            model.add(transform_layer)
+
+        return model
+
+
+class HyperImageRandAugment(HyperImageAugment):
```
omalleyt12, line 174: I think instead of having `HyperImageRandAugment` and `HyperImageFixedAugment`, it would be better to have one `HyperImageAugment` class with an optional argument `rangaug_count` that if not set, always applies all transforms

That way users don't have to choose b/t objects, and we can just provide the defaults that we think are best.

WDYT?

kerastuner/applications/augment.py
```diff
@@ -0,0 +1,303 @@
...
+            minimum factor is 0; if None, the transform is disabled. All factors
+            are within [0, 1].
+    """
+    def __init__(self,
+                 input_shape=None,
+                 input_tensor=None,
+                 rotate=0.2,
+                 translate_x=0.1,
+                 translate_y=0.1,
+                 contrast=0.1,
```
omalleyt12, line 74: Can we add an argument like `max_augmentations=None` to control randomly selecting a few augmentations or running all augs.

kerastuner/applications/augment.py
```diff
@@ -0,0 +1,303 @@
...
+
+    # Arguments:
+        input_shape: Optional shape tuple, e.g. `(256, 256, 3)`.
+            One of `input_shape` or `input_tensor` must be
+            specified.
+        input_tensor: Optional Keras tensor (i.e. output of
+            `layers.Input()`) to use as image input for the model.
+            One of `input_shape` or `input_tensor` must be
+            specified.
+        randaug_mag: Number or list of 2 numbers. Default is 1. It
```
omalleyt12, line 192: Let's remove this and rely on per-augmentation hyperparameters

PR #365, opened 2020-07-21
A FineTuner class that automatically unfreezes

yixingfu:
FineTuner class that automatically run trial for two times, while unfreezing a part of the model between these two. The FineTuner implementation is suggested by @Aviously . This is important for Hypermodels that relay on pretrained weights, such as EfficientNet (hypermodel in PR #363 ). When used with a model that does not naturally start with a part of the layers unfrozen, this tuner roughly works like original tuner.

To consistently unfreeze any model (especially thinking of efficientnet, resnet etc that are composed of several blocks) such that the boundary between frozen/unfrozen layers does not break the block structure, I implemented a 'PseudoSequential' class that does the partition. The idea is similar to finding articulation point in undirected graph. The partition that happen only at articulation points will respect block structure. 

What I am strongly uncertain and need suggestion is where I should put these contribution. Right now I have it under tuner/FineTuner.py. Maybe it is better as an application? Or maybe as part of tuner or multi_execution_tuner under engine? Or as a special tutorial? What do you think? @omalleyt12 

Also @haifeng-jin, not sure maybe this should better go to autokeras instead of keras-tuner?

### DIFF
--- a/kerastuner/tuners/finetuner.py
+++ b/kerastuner/tuners/finetuner.py
@@ -1,0 +1,295 @@
+from tensorflow import keras
+from tensorflow.keras import layers
+from ..engine import hyperparameters as hp_module
+from ..engine import multi_execution_tuner
+from ..engine import oracle as oracle_module
+from ..engine import trial as trial_lib
+
+class FineTuner(multi_execution_tuner.MultiExecutionTuner):
+    """FineTuner automatically transfer-learn and fine-tunes a hypermodel that
+       start with pretrained weights.
+
+    # Arguments:
+        oracle: Instance of Oracle class.
+        hypermodel: Instance of HyperModel class (or callable that takes
+            hyperparameters and returns a Model instance).
+            This hypermodel should: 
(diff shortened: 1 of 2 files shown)


haifeng-jin:
@yixingfu I think is should stay in Keras Tuner. Otherwise the Keras Tuner users cannot use it. AutoKeras can always override the class and use it in AutoKeras.

kerastuner/tuners/finetuner.py
```diff
@@ -0,0 +1,295 @@
+from tensorflow import keras
+from tensorflow.keras import layers
+from ..engine import hyperparameters as hp_module
+from ..engine import multi_execution_tuner
+from ..engine import oracle as oracle_module
+from ..engine import trial as trial_lib
+
+class FineTuner(multi_execution_tuner.MultiExecutionTuner):
```
omalleyt12, line 8: For this PR, let's wait on the outcome of the design doc on how to best handle transfer learning

PR #379, opened 2020-08-12
single out training loop to a _run_fit method

yixingfu:
Put the `model.fit()` part of tuners into a separate method `_run_fit`, so that `run_trial` will call this method. This change makes easier to override a tuner if we only need to change the training loop. For example, if we want to override `MultiExecutionTuner` to modify training loop, we need to still include everything that is not related to the training loop itself.

### DIFF
--- a/kerastuner/engine/base_tuner.py
+++ b/kerastuner/engine/base_tuner.py
@@ -147,7 +147,7 @@ class BaseTuner(stateful.Stateful):
         ```python
         def run_trial(self, trial, x, y, val_x, val_y):
             model = self.hypermodel.build(trial.hyperparameters)
-            model.fit(x, y)
+            self._run_fit(model, trial.hyperparameters, x, y, val_x, val_y)
             loss = model.evaluate(val_x, val_y)
             self.oracle.update_trial(
               trial.trial_id, {'loss': loss})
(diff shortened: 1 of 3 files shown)


kerastuner/engine/base_tuner.py
```diff
@@ -147,7 +147,7 @@ def run_trial(self, trial, *fit_args, **fit_kwargs):
         ```python
         def run_trial(self, trial, x, y, val_x, val_y):
-            model.fit(x, y)
```
omalleyt12, line 150: For our example, let's still use `model.fit`. We want to keep things as simple as possible in the user's minds

kerastuner/engine/base_tuner.py
```diff
@@ -163,6 +163,38 @@ def run_trial(self, trial, x, y, val_x, val_y):
         """
         raise NotImplementedError
 
+    def _run_fit(self, model, hp, *args, **kwargs):
```
omalleyt12, line 166: Let's put this in Tuner, not BaseTuner, since BaseTuner is supposed to be generic and work for non-Keras models

kerastuner/engine/tuner.py
```diff
@@ -124,6 +124,9 @@ def run_trial(self, trial, *fit_args, **fit_kwargs):
         This method is called during `search` to evaluate a set of
         hyperparameters.
 
+        The training loop should be in `_run_fit` method for easier
```
omalleyt12, line 127: Since this is private method, we shouldn't mention it in the docstring

kerastuner/engine/base_tuner.py
yixingfu: Thanks for the comment. I'll make the change

PR #382, opened 2020-08-17
Formatting float to avoid showing too many decimals

yixingfu:
According to suggestion from @fchollet, avoid displaying full length of floats.


yixingfu:
duplicate.

Issue #381, opened 2020-08-13
Handling division by zero


SHASHANK-PRO-05:
https://github.com/keras-team/keras-tuner/blob/a17321428a55975f55acb5b9270d6d73ea11dcc9/kerastuner/engine/hyperparameters.py#L1064

Hey team,
There should be the handling of division by zero here.

yixingfu:
@haifeng-jin @omalleyt12 
I can think of a few solutions of this:
1. automatically turn anything with same min/max to `Fixed`
2. since this is only used in bayesian, we can make `_nonfixed_space` in bayesian tuner to recognize it as fixed space
3. assign 1 (or any number) to this
What do you think?

haifeng-jin:
@yixingfu I think the third solution would be the best.
For the first one, it would confuse the user if we change the hp type under the hood.
For the second one, we may use in other tuners in the future as well, not only bayesian.

Thanks.

yixingfu:
@haifeng-jin Thanks for the input. BTW should we also check `value` to be between min and max here?

haifeng-jin:
@yixingfu That would be even better. The check should be done when creating the hyperparam.

yixingfu:
> @yixingfu That would be even better. The check should be done when creating the hyperparam.

That is still not safe, Bayesian oracle may potentially create a trial with value out of bound (or am I missing anything?) and these values are set into `hps.values` without going through hyperparam creation.

yixingfu:
I'd say we can throw an error here if value is not in bound, wdyt

PiotrAntoniak:
In case if someone is wondering if it has been fixed - it just happened to me a second ago. However, overwriting  the project fixes that.

PR #389, opened 2020-08-20
handle division by zero

yixingfu:
#381 ;
also check value is in bound when converting to probability.

### DIFF
--- a/kerastuner/engine/hyperparameters.py
+++ b/kerastuner/engine/hyperparameters.py
@@ -1061,9 +1061,10 @@ def value_to_cumulative_prob(value, hp):
         # Center the value in its probability bucket.
         return (index + 0.5) * ele_prob
     elif isinstance(hp, (Int, Float)):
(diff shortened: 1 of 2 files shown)


kerastuner/engine/hyperparameters.py
```diff
@@ -1054,11 +1054,18 @@ def value_to_cumulative_prob(value, hp):
...
+        if not hp.values:
+            return 0
         ele_prob = 1 / len(hp.values)
         index = hp.values.index(value)
         # Center the value in its probability bucket.
         return (index + 0.5) * ele_prob
     elif isinstance(hp, (Int, Float)):
+        if value < hp.min_value or value > hp.max_value:
+            raise ValueError('`value` must be in the range defined in the '
+                             'HyperParameter.')
```
haifeng-jin, line 1066: I think this check is already done at the sampling function.
We should assume the value is in the range.

haifeng-jin:
This pr is stale, use #586 .

PR #399, opened 2020-09-02
Add _on_build_begin to tuner for AutoKeras to override

haifeng-jin:
AutoKeras will need to use tf.data preprocessing pipelines to transform the training data and get the input shape of the Keras model.
So added _on_build_begin function before building the hypermodel in the tuner, and pass the fit_args as a dictionary by reference for modification.


fchollet (approved):

omalleyt12 (approved):
Thanks for the PR! LGTM

PR #414, opened 2020-10-05
cleanup the override functions for ak

haifeng-jin:
Removed `_on_build_begin` and `_on_train_begin`.
Extracted the build and fit process as a separate function for `Tuner` and `MultiExecutionTuner` to reuse,
and for AutoKeras to override.

The refactor is to support the adaptive batch size feature in AutoKeras, which needs to surround the fit function with a try and except for OOM.


omalleyt12 (approved):
Thanks for the PR!

PR #423, opened 2020-10-17
HyperParameters get now raises KeyError if _hps doesn't contain key.

### LINKED ISSUE #411: HyperParameters.get(): should raise KeyError rather than ValueError when name is not Found
Let ``hp`` be an empty ``HyperParameters`` container. When calling ``hp.get('anything')`` or ``hp['anything']``, a ``ValueError`` is raised. It would be more pythonic if it raised a ``KeyError``.

``HyperParameters`` objects are ``dict``-like containers . A ``dict`` would raise a ``KeyError``.

*keras-tuner version: 1.0.1*

jpodivin:
resolves #411

The relevant method (get of HyperParameters) raised ValueError in two cases. 

1. If the key wasn't in values
2. If the key wasn't also in _hps

Since only the second case actually indicates non-existence of the hyperparameter by that name, I believe only it should be altered. as it is in this PR.

### DIFF
--- a/kerastuner/engine/hyperparameters.py
+++ b/kerastuner/engine/hyperparameters.py
@@ -659,7 +659,7 @@ class HyperParameters(object):
         try:
             self.get(name)
             return True
-        except ValueError:
+        except (KeyError, ValueError):
             return False
 
     def Choice(self,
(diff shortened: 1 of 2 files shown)


jpodivin:
I figured out what was wrong with test_dict_methods. But I'm still unsure as to why bayesian_test.py::test_distributed_optimization keeps failing.

jpodivin:
Tests finally came up all clear.

haifeng-jin (approved):
Thanks for the PR! LGTM

PR #424, opened 2020-10-18
correctly get best epoch during save checkpoints

haifeng-jin:


fchollet (approved):

omalleyt12 (approved):
Thanks for the PR! LGTM

PR #428, opened 2020-10-24
Implemented check on upper bound of the hyperparameter values, so it …

jpodivin:
…is always greater than the lower bound.


haifeng-jin (approved):
Thanks for the PR! LGTM

PR #418, opened 2020-10-13
Integrate with TensorBoard HParams plugin

chongyouquan:
- Add HParams KerasCallback automatically if TensorBoard callback is present
- Add converter from KerasTuner Hyperparameters to TensorBoard HParams
- Extract the TensorBoard logging directory to _get_tensorboard_dir
- Fix typos

PR #432, opened 2020-10-27
Add missing self

carlthome:
Unless there's some runtime magic that I'm missing, `_get_tensorboard_dir` isn't defined at this point so I assume the intention was to call the class method here.


carlthome:
@chongyouquan, could confirm perhaps?

https://github.com/keras-team/keras-tuner/pull/418/files#diff-a9b15c6772daf638260091411f939f369e5ff4ee505f5cf65659039c7413e31aR285

chongyouquan:
Yes, good catch, thanks!

haifeng-jin (approved):
Thanks for the PR! LGTM

PR #433, opened 2020-10-27
Fix missing self and modify tests

### LINKED ISSUE #142: Feature Request: Using tensorboard hparams.
Curently I didn't see a good way to save and load the hyperparameters to files.
Additionally, there is already an HParams thing in tensorflow see [here](https://www.tensorflow.org/tensorboard/hyperparameter_tuning_with_hparams)

Would be nice if both were integrated or a function to convert between both. 
I would post some coding examples, but the link above speaks in volumes.

chongyouquan:
- Add missing self in _configure_tensorboard_dir
- Use arange instead of linspace to match hyperparameters.py
- Deflake multifloat test
- Fixes #142

### DIFF
--- a/kerastuner/engine/tuner.py
+++ b/kerastuner/engine/tuner.py
@@ -18,7 +18,6 @@ from __future__ import division
 from __future__ import print_function
 
 import copy
-import numpy as np
 import os
 
 from tensorboard.plugins.hparams import api as hparams_api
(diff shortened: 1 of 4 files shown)


chongyouquan:
@omalleyt12 Could you help me review this? Thanks!

omalleyt12 (approved):
Thanks for the PR! LGTM

PR #434, opened 2020-10-30
fix typo in Tuner class description

alcantarar:
the `tf.distribute.MirroredStrategy` example has missing apostrophes. Example now matches https://www.tensorflow.org/api_docs/python/tf/distribute/MirroredStrategy


haifeng-jin (approved):
Thanks for the PR! LGTM

PR #437, opened 2020-11-08
Altered .gitignore

jpodivin:
VS Code offers rather nice option of developing inside of a generated container that needs only slight customization. 
Unfortunately, the generated files are still considered candidates for tracking by git.

Not with this addition however.


haifeng-jin (approved):
Thanks for the PR! LGTM. It is also good for GitHub Codespace.

Issue #448, opened 2020-11-23
Issue since upgrade 1.0.2 (module 'tensorflow.keras.layers.experimental.preprocessing' has no attribute 'RandomRotation')


nidragedd:
Hi,

Since Nov the 21th I have issues with some of my jobs in production. I digged and found that this is due to keras-1.0.2 upgrade (before I was using 1.0.1).
Maybe a brad practice but in my requirements.txt, versions were not frozen so it has upgraded by itself this WE.

Anyway, what was previously working like a charm has now this error:
```
python main_predict.py --environment e--job_type j
Traceback (most recent call last):
  File "main_predict.py", line 11, in <module>
    from src.modeling.rnn.model.rnn import RNN
  File "/home/ec2-user/workspace/user/predict_rnn_van/src/modeling/rnn/model/rnn.py", line 12, in <module>
    import kerastuner as kt
  File "/home/ec2-user/anaconda3/envs/tensorflow2_p36/lib/python3.6/site-packages/kerastuner/__init__.py", line 19, in <module>
    from . import applications
  File "/home/ec2-user/anaconda3/envs/tensorflow2_p36/lib/python3.6/site-packages/kerastuner/applications/__init__.py", line 19, in <module>
    from .augment import HyperImageAugment
  File "/home/ec2-user/anaconda3/envs/tensorflow2_p36/lib/python3.6/site-packages/kerastuner/applications/augment.py", line 34, in <module>
    'rotate': preprocessing.RandomRotation,
AttributeError: module 'tensorflow.keras.layers.experimental.preprocessing' has no attribute 'RandomRotation'
```

I rollbacked to 1.0.1 and it works but I am letting you know this because the upgrade is not seamless.

tensorflow2_p36 has tensorflow 2.1.2 and python 3.6 on UNIX machines.

informatica92:
Same issue here. Solution: rollback to 1.0.1 version

Ab-Abdurrahman:
rollback to keras-tuner 1.0.1 version did not help. Instead, I updated TensorFlow to 2.2 from 2.1 that I was using previously

Ab-Abdurrahman:
rollback to keras-tuner 1.0.1 version did not help. Instead, I updated TensorFlow to 2.2 from 2.1 that I was using previously

Leon-Brant:
My TensorFlow-GPU version is the latest 2.6, but I still work with

PR #451, opened 2020-11-29
Update tensorflow requirement to 2.2

gqqnbig:
[augment.py](https://github.com/keras-team/keras-tuner/blob/master/kerastuner/applications/augment.py#L34) references `experimental.preprocessing.RandomRotation`, which, according to #448, is available since tf 2.2.

If keras-tuner 1.0.3 is installed with tf 2.1, `import kerastuner` throws exceptions right away.


haifeng-jin:
We will try to support TF 2.0 by fixing the import issues.

Thanks for your PR!

PR #467, opened 2020-12-23
Enable the tuning of scikit-learn pipeline

qingquansong:
Enable the tuning of scikit-learn pipeline.

Test with unit test & integration test.


haifeng-jin:
Use a new pr with a new branch in #478

PR #475, opened 2021-01-28
Changes in the gradient tape block

ariG23498:
The proposed change asks to indent the gradient outside the gradient tape block.


haifeng-jin (approved):
Thanks for the PR! LGTM

PR #505, opened 2021-03-11
remove some scipy usages

haifeng-jin:
This PR removes all the scipy usages except for the L-BFGS-B function minimizing algorithm.
For this use case, numpy is faster than TF, so using numpy to implement.
For TF implementation, refer to #506.


fchollet (approved):
LGTM, thank you for the PR!

PR #516, opened 2021-04-04
Unnecessary explicit object inheritance.

jpodivin:
Minor stylistic fix. As all classes implicitly
inherit from object in Python 3, explicit inheritance
from object is no longer necessary.

Signed-off-by: Jiri Podivin <jpodivin@gmail.com>


haifeng-jin:
We would keep the object extension for now since some dependencies are still using Python 2.
It would be easier for them to rebase and maintain a fork to continue the support.

jpodivin:
Understood. Closing PR.

PR #517, opened 2021-04-04
Condition now checks for both NaNs and Infs

jpodivin:
Original condition only checked for NaNs,
but allowed Infs, both positive and negative,
to slip by. This patch unifies checks for
Infs and NaNs using numpy.isfinite, which
evaluates true for all values, except for infinity
and NaN.

Signed-off-by: Jiri Podivin <jpodivin@gmail.com>


haifeng-jin (approved):
Thanks for the PR!

PR #518, opened 2021-04-07
Update tutorial

mkg33:
The typo has already been corrected in the documentation but it could also be corrected manually without autogenerating the whole documentation again.


mkg33:
@googlebot I signed it!

mkg33:
Actually, I'm not entirely sure it makes sense to do it this way. I suppose the website should get rebuilt from source because there are several updates that are already in the source but not in the stable online documentation.

haifeng-jin (approved):
Thanks for the PR! Since rebuilding the website will take while to happen in the future, this PR can be merged as a temporary solution.

mkg33:
@haifeng-jin Good to know!

PR #543, opened 2021-06-09
enable importing oracles with their real class name

### LINKED ISSUE #541: Change the Oracle names from RandomSearch to RandomSearchOracle
While changing, should keep the original import statements just not to break people.
The solution should be as simple as some new import statements which import the classes directly without changing the names.

haifeng-jin:
resolve #541


fchollet (approved):
Thanks for the PR. After merging / releasing we should also update the names documented at keras.io.

PR #557, opened 2021-06-17
removing references to CloudLogger in the docs

haifeng-jin:
The `CloudLogger` is removed from the docs.
We keep the implementation of `CloudLogger` just not to break people and may use it in the future.


fchollet (approved):
LGTM, thanks!

PR #562, opened 2021-06-25
Revise is_active function definition

haifeng-jin:
Consolidate the functionality of `is_active` to checking the conditions.
Decouple it from whether the values contain the hp or not.
Otherwise, it enforces the `Oracle` to only generate values for active hps.

After this patch, the user can choose to let their custom oracle generate values for inactive values, which won't cause any problem.
It would not affect the feature of "always returning None when using a inactive hp in the build_model function" since this feature is implemented independent to the `.values` attributes, and independent to the `is_active` function.


fchollet (approved):
Thanks for the PR! LGTM. Please pull copyedits first.

Issue #549, opened 2021-06-14
Continuing the search does not update the trial number in the log


LucaUrbinati44:
When I want to continue an hyperparameter search with `tuner.search()`, the log that is printed as output starts from 1 instead from the number of the last trial + 1.
For example, I arrived at Trial 23, but after initializing the tuner with the same `directory` and `project`, it says: `Search: Running Trial #1`, instead of `Search: Running Trial #24`. 
However, the tuner is able to load the best trial among the last 23, so the state of the searching process is preserved.
This is a minor issue, but I wanted to let you know. Thanks

vardhanaleti:
Hello @haifeng-jin, I would like to work on this issue. 

I've gone through the source code and here's how I feel I can start ( Its one of the first issues I'm taking a look at so please don't mind me asking a few questions about it). 

As the _set_project_dir() function sets the directory and reloads the oracle, the number of previously stored trials can be saved and be used as the trial_number in the display function rather than reinitializing every time the tuner is initialized. 

Thank you.

haifeng-jin:
@vardhanaleti Thank you for taking this issue.

I haven't really dig into the cause of this bug yet.
But for it to work as expected, we should be able to get the number trials from a loaded oracle and the display function should just accept that normally from the oracle (directly or indirectly) as normal.

vardhanaleti:
Absolutely, I've edited and ran it accordingly and it works as expected. 
Before making a PR, I'm going through the contribution guidelines and setting the right environment for testing as I'm new to making contributions.
Thank you.

PR #595, opened 2021-08-25
Fix the trial number update issue #549  

vardhanaleti:
The trial number is now initialized to the previous number of trials done. This the correct trail number to get logged when oracle is reloaded.


vardhanaleti:
> Thanks for the PR!

Thank you! This being my first fix, I got to learn how tests and the whole contribution process works 😃

PR #601, opened 2021-09-03
update metrics docstrings

haifeng-jin:


fchollet (approved):
Thanks for the PR! LGTM -- added some copyedits.

PR #602, opened 2021-09-03
Change MultiExecutionTuner to Tuner

haifeng-jin:
The original `Tuner` class is for single execution per trial and reporting every epoch to the oracle.
The `MultiExecutionTuner` runs multiple executions per trial and reports to the oracle only once per trial.
All built-in tuners are `MultiExecutionTuner` subclasses.

Users thought the code in `Tuner` defines the behavior of the subclass tuners.
However, the functions are actually overridden by `MultiExecutionTuner` and results in different behaviors, which confuses the user.
The single execution feature in the `Tuner` class is for internal use only, and not called in the OSS code at all.

To resolve the confusion, we swapped the two classes.
Changed `MultiExecutionTuner` to `Tuner`, and the original `Tuner` to `SingleExecutionTuner`.
Now `SingleExecutionTuner` is a subclass of the `Tuner`.
It will be removed in the OSS and move to internal in the future.

We kept the `MultiExecutionTuner` class for backward compatibility.

### DIFF
--- a/keras_tuner/engine/multi_execution_tuner.py
+++ b/keras_tuner/engine/multi_execution_tuner.py
@@ -12,115 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"Tuner that runs multiple executions per Trial."
-
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
-
-import collections
-import copy
-import os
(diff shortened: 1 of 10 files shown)


fchollet (approved):
LGTM, thanks for the PR!

PR #607, opened 2021-09-10
fix missing pandas

howl-anderson:
fix missing pandas

Error message :
```
.../keras_tuner/tuners/sklearn_tuner.py:21: in <module>
    import pandas as pd
E   ModuleNotFoundError: No module named 'pandas'
```

### DIFF
--- a/keras_tuner/tuners/sklearn_tuner.py
+++ b/keras_tuner/tuners/sklearn_tuner.py
@@ -18,8 +18,12 @@ import pickle
 import warnings
 
 import numpy as np
-import pandas as pd
 import tensorflow as tf
+
+try:
+    import pandas as pd  # pytype: disable=import-error
+except ImportError:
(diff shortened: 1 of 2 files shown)


haifeng-jin:
@howl-anderson Thank you for the fix!
We would prefer fixing it in the following way:
```py
try:
    import pandas as pd
except ImportError:
    pd = None
```

And change https://github.com/keras-team/keras-tuner/blob/master/keras_tuner/tuners/sklearn_tuner.py#L37
to
```py
elif pd is not None and isinstance(data, pd.DataFrame):
```

That is the only place pandas is used.

howl-anderson:
@haifeng-jin Thank you for your review. I will update this PR to make changes accordingly.

howl-anderson:
@haifeng-jin Thanks for update the setup.py

PR #611, opened 2021-09-21
remove single_execution_tuner

haifeng-jin:
The SingleExecutionTuner is moved to internal use only.


fchollet (approved):
LGTM, thanks for the PR!

PR #623, opened 2021-10-12
Remove KerasHyperModel

haifeng-jin:
Refactored to remove `KerasHyperModel`.
The logic is moved into `Tuner`.

Before refactoring, `Tuner.hypermodel` is not the `HyperModel` user provided, but a `KerasHyperModel`, which wraps the original `HyperModel`.
It confuses the user when accessing `tuner.hypermodel` or using `self.hypermodel` when overriding any method in `Tuner`.

`KerasHyperModel` is to:
* Build model multiple times on build failures.
* Override the `compile()` args.
* Put model building and compiling into `distribution_strategy` scope.

These logic of managing the model without the hps belongs to `Tuner`.


fchollet (approved):
LGTM, thanks for the PR

PR #626, opened 2021-11-01
bug fix: override compile args when loading model

haifeng-jin:
The bug:

When loading the best model, the `loss`, `optimizer` and `metrics` passed to the tuner are not used. It loads the compile args used in the hypermodel.

Fix:

Minor refactor. Whenever trying to build the model in `Tuner`, just call `_try_build()`, which will build and override the compile args.


fchollet (approved):
LGTM, thanks for the PR!

PR #634, opened 2021-11-24
Make hyperparameters names visible in Display output

C-Pro:
Moved Hyperparameter name to the last column and removed length description as discussed in #491

Result looks like this:
```
Search: Running Trial #1

Value             |Best Value So Far |Hyperparameter
False             |False             |structured_data_block_1/normalize
False             |True              |structured_data_block_1/dense_block_1/use_batchnorm
3                 |3                 |structured_data_block_1/dense_block_1/num_layers
512               |512               |structured_data_block_1/dense_block_1/units_0
0                 |0                 |structured_data_block_1/dense_block_1/dropout
512               |512               |structured_data_block_1/dense_block_1/units_1
0                 |0                 |classification_head_1/dropout
adam_weight_decay |adam_weight_decay |optimizer
0.001             |0.001             |learning_rate
128               |128               |structured_data_block_1/dense_block_1/units_2
```

### DIFF
--- a/keras_tuner/engine/tuner_utils.py
+++ b/keras_tuner/engine/tuner_utils.py
@@ -142,14 +142,14 @@ class Display(object):
             print("Total elapsed time: {}".format(time_elapsed_str))
 
     def show_hyperparameter_table(self, trial):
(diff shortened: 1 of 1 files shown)


haifeng-jin (approved):
Thanks for the PR! This is a long expected feature.

PR #650, opened 2022-01-29
Ensure hallucination checks if the Gaussian regressor has been fit be…

brydon:
…fore calling predict in the Bayesian oracle.

Previously the oracle would attempt to call the predict method of the Gaussian regressor without checking if the regressor has been fit yet. In distributed runs, this could lead to a cryptic error about _x_train if the Bayesian oracle attempted to hallucinate on a trial recently initialized by a different worker process.

`raise _InactiveRpcError(state) grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with: status = StatusCode.UNKNOWN details = "Exception calling application: 'GaussianProcessRegressor' object has no attribute '_x_train'" debug_error_string = "{"created":"@1643154861.118197004","description":"Error received from peer ipv4:127.0.0.1:8000","file":"src/core/lib/surface/call.cc","file_line":1074,"grpc_message":"Exception calling application: 'GaussianProcessRegressor' object has no attribute '_x_train'","grpc_status":2}"`

This error is especially common when hyperparameter optimizing with a long-running, training loop. Now, the Bayesian oracle ensures that the Gaussian regressor has been previously fit before it attempts to use that regressor to predict for "hallucinations".

### DIFF
--- a/keras_tuner/engine/tuner_utils_test.py
+++ b/keras_tuner/engine/tuner_utils_test.py
@@ -92,14 +92,11 @@ def test_convert_to_metrics_with_float():
 
 
 def test_convert_to_metrics_with_dict():
-    assert (
-        tuner_utils.convert_to_metrics_dict(
-            {"loss": 0.2, "val_loss": 0.1},
-            obj_module.Objective("val_loss", "min"),
-            "func_name",
-        )
-        == {"loss": 0.2, "val_loss": 0.1}
-    )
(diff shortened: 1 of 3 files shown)


haifeng-jin (approved):
Thank you for the PR!

haifeng-jin:
@brydon You will need to run `shell/format.sh` to format the code before I can merge it. Thanks.

brydon:
Formatting fixed!

Issue #605, opened 2021-09-07
Have a more human readable string as the trial_id.


haifeng-jin:

JamesCallanan:
I don't think id's need to be human readable. 

The most important thing for an id should be its ability to be used to identify something. 

By giving trials from different searches the same ids, they become far less useful as identifiers (as they are not unique). As a result, I think this change doesn't make much sense.

PR #653, opened 2022-01-30
Resolves #652 and #605: Make human readable trial_id and sync trial numbers between worker Displays

### LINKED ISSUE #652: When running distributed, ongoing count of trial number is not synced between workers
When running keras_tuner distributed, the `Display` prints out a string like `Search: Running Trial #x`.
Initially, these numbers make sense. Meaning, if I run 5 workers I'll see `Search: Running Trial #1` in the first, `Search: Running Trial #2`, all the way to `Search: Running Trial #5` in the fifth. However, the second iteration of the first worker will read `Search: Running Trial #2` as well (despite the fact that the "second trial" was already run). 

This is purely a cosmetic bug -- that is `max_trials` is always respected. However, there is no easy way to check the progress of the workers except by doing some kind of count of the `trial_xxxx` folders in the output directory.

**To Reproduce**
The following MWE is composed of three files.
Here is `trial_id_mwe.py` - a simple example of using `keras_tuner` to maximize a Gaussian.
```
import keras_tuner as kt
import numpy as np
import time


class MWETuner(kt.Tuner):
    def run_trial(self, trial):
        hp = trial.hyperparameters
        a = hp.Float('a', -2, 2)
        time.sleep(10) # Simulate training loop
        
        return {"score": np.exp(-a**2)}


if __name__ == "__main__":
    oracle = kt.oracles.RandomSearchOracle(
        objective=kt.Objective("score", "max"),
        max_trials=6
    )
    
    tuner = MWETuner(
        oracle=oracle,
        directory="results",
        project_name="mwe"
    )
    
    tuner.search()
```
I then launch my chief with this bash script `run_chief.sh`
```
export KERASTUNER_TUNER_ID="chief"
export KERASTUNER_ORACLE_IP="127.0.0.1"
export KERASTUNER_ORACLE_PORT="8000"
export TF_CPP_MIN_LOG_LEVEL="3"
export CUDA_DEVICE_ORDER=""
export CUDA_VISIBLE_DEVICES=""

python trial_id_mwe.py
```

and my two worker classes with this bash script `run_worker.sh`
```
export KERASTUNER_TUNER_ID="tuner$1"
export KERASTUNER_ORACLE_IP="127.0.0.1"
export KERASTUNER_ORACLE_PORT="8000"
export TF_CPP_MIN_LOG_LEVEL="3"
export CUDA_DEVICE_ORDER=""
export CUDA_VISIBLE_DEVICES=""

python trial_id_mwe.py
```
Where I launch `sh run_worker.sh 0` and `sh run_worker.sh 1` for the two tuners. Note: I disable GPU to make this example run easily on CPU, though this error persists when training over multiple GPUs as well.

**Expected behavior**
Would expect the number to sync between worker `Oracle`s.

**Would you like to help us fix it?**
Pull request submitted.

brydon:
By allowing the `Oracle` class to assign the `trial_id` based on the current size of the `trials` list, the trial_id becomes more humanly readable (re: #605). 

Moreover, the trial number is now intrinsically a part of the `trial` object, meaning that distributed worker Oracles can display the correct trial number being considered without the need of communicating directly to the chief over the socket. By leveraging the chief oracle's original distribution strategy, the worker `Display`s are kept in sync and the current trial number does not become decoupled over successive iterations (re: #652).

As an example, here is the result of running the MWE code from #652 on the current `master` branch of `keras_tuner`:
mwe_before
the worker in the bottom left was run first and correctly called the first trial `Trial 1`. The worker in the bottom right was run second and correctly labelled its first trial as `Trial 3` (as the first worker had already completed 2 iterations at this point). However, the next run of the first `tuner` was also labelled Trial 3 and all Tuners finished after 6 total iterations had been completed, but judging from the output no `Trial 5` or `Trial 6` was completed (though all trials _were_ completed in actuality, as evidenced by the `ls` command run in the top terminal after all workers concluded).

In contrast, here is the result of running the same MWE code from #652 but using this updated branch of `keras_tuner`
mwe_fixed
we see each trial uniquely represented in the worker `Display`s without spurious double-counting. Also, the `trial_id` folder names are human-readable in the final `ls` command in the top terminal.

Moreover, the `trial_id` strings are padded with the appropriate number of zeros (as dictated by `max_trials`) so that trials are displayed in the correct order in a file directory for larger distributed runs.
mwe_3

### DIFF
--- a/keras_tuner/engine/oracle.py
+++ b/keras_tuner/engine/oracle.py
@@ -183,7 +183,9 @@ class Oracle(stateful.Stateful):
         if tuner_id in self.ongoing_trials:
             return self.ongoing_trials[tuner_id]
 
-        trial_id = trial_lib.generate_trial_id()
+        # Make the trial_id the current number of trial, pre-padded with 0s
+        trial_id = "{{:0{}d}}".format(len(str(self.max_trials)))
+        trial_id = trial_id.format(len(self.trials))
 
         if self.max_trials and len(self.trials) >= self.max_trials:
             status = trial_lib.TrialStatus.STOPPED
--- a/keras_tuner/engine/tuner_utils.py
+++ b/keras_tuner/engine/tuner_utils.py
@@ -95,7 +95,6 @@ class Display(object):
     def __init__(self, oracle, verbose=1):
         self.verbose = verbose
         self.oracle = oracle
-        self.trial_number = len(self.oracle.trials)
         self.col_width = 18
 
         # Start time for the overall search
@@ -107,7 +106,7 @@ class Display(object):
     def on_trial_begin(self, trial):
         if self.verbose >= 1:
 
-            self.trial_number += 1
+            self.trial_number = int(trial.trial_id) + 1
             print()
             print("Search: Running Trial #{}".format(self.trial_number))
             print()
--- a/keras_tuner/engine/tuner_utils_test.py
+++ b/keras_tuner/engine/tuner_utils_test.py
@@ -92,14 +92,11 @@ def test_convert_to_metrics_with_float():
 
 
 def test_convert_to_metrics_with_dict():
-    assert (
-        tuner_utils.convert_to_metrics_dict(
-            {"loss": 0.2, "val_loss": 0.1},
-            obj_module.Objective("val_loss", "min"),
-            "func_name",
-        )
-        == {"loss": 0.2, "val_loss": 0.1}
-    )
+    assert tuner_utils.convert_to_metrics_dict(
+        {"loss": 0.2, "val_loss": 0.1},
+        obj_module.Objective("val_loss", "min"),
+        "func_name",
+    ) == {"loss": 0.2, "val_loss": 0.1}
 
 
 def test_convert_to_metrics_with_list_of_floats():
--- a/keras_tuner/tuners/bayesian.py
+++ b/keras_tuner/tuners/bayesian.py
@@ -50,7 +50,7 @@ def matern_kernel(x, y=None):
     # nu = 2.5
     dists = cdist(x, y)
     dists *= math.sqrt(5)
-    kernel_matrix = (1.0 + dists + dists ** 2 / 3.0) * np.exp(-dists)
+    kernel_matrix = (1.0 + dists + dists**2 / 3.0) * np.exp(-dists)
     return kernel_matrix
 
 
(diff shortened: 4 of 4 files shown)


haifeng-jin (approved):
Thanks for the PR! I checked its save and load mechanism. It is safe since the trials are saved directly to the hard drive, refer to `Oracle.reload()`.

haifeng-jin:
@brydon You will need to run `shell/format.sh` to format the code before I can merge it. Thanks.

brydon:
Formatting fixed!

JamesCallanan:
I don't think id's need to be human readable. 

The most important thing for an id should be its ability to be used to identify something. 

By giving trials from different searches the same ids, they become far less useful as identifiers (as they are not unique). As a result, I think this change doesn't make much sense.

brydon:
For different searches the trials are stored in different directories, so the trials will still be uniquely identifiable. For instance,
`search_1/trial_001/` differs from `search_2/trial_001/`.

JamesCallanan:
I know you won't completely lose track of the trial this way but it's sub optimal if an identifier can only be used as an identifiers alongside extra information (the extra information being the search it is linked to in this case).

if you were trying to store trials in one place e.g. one database table (which is a logical thing to do). It would make sense that the trial_id could be used as a primary key.

brydon:
In terms of the use-case you described, storing different trials from different searches in the _same_ database, I don't think I agree that that is a sufficiently general design choice (suppose one search is for HP tuning an imaging classifier and the other for HP tuning an NLP sentiment analyzer, why would I need both in the same database?). So in the edge cases you describe, I think I would suggest using a hash of the full string (`search_1/trial001`, for instance) as the primary key. 

In terms of needing to know which trial was first, second, or 34th: this can be very useful information, in my view, in at least two different ways.
1. For long-running tuning processes, this can be useful for checking the progress of the tuner. 
2. In a _post-hoc_ sense it can be useful to plot the score vs. trial number to examine the convergence of the tuner (for instance, for the Bayesian optimizer). In such a case, one would expect the score to saturate as the trial number increases. For a random tuner, this is less useful. 

If others disagree, and instead think preserving the pseudo-uniqueness of trial IDs across different searches is worthwhile, I'd instead propose a middle-ground solution where the trial id takes the form `trial_001_HEXSTRING` and would be happy to implement this.

JamesCallanan:
1. Are you be able to check the progress while a search is running? Would be interested to know how, that sounds v. useful.

2. Didn't think of the use case for the Bayesian optimisation example. Could be useful there.

What I'm doing is carrying out several hyperparameter searches for one classification task and looking to store summaries of all relevant trials in one place. Suspect that isn't an edge case. But I guess we can wait to see if others have the same issue.

haifeng-jin:
I have replied in #669 .

Thanks.

PR #672, opened 2022-03-24
add --profile=black to isort

LukeWood:
This causes isort to respect the rules of black.  Without this,
if you have any long paths isort will wrap them as follows:

from my_long_path_import_something.something import (
  Symbol
)

while black does:

from my_long_path_import_something.something import (
  Symbol
)


haifeng-jin (approved):
Awesome!

PR #684, opened 2022-04-25
Adopt safe model and trial saving practices in the multi-worker setting

jamesmullenbach:
Based on the [reference guide](https://www.tensorflow.org/tutorials/distribute/multi_worker_with_keras#model_saving_and_loading) for saving models when training with multiple workers.

### DIFF
--- a/keras_tuner/engine/oracle.py
+++ b/keras_tuner/engine/oracle.py
@@ -114,7 +114,7 @@ class Oracle(stateful.Stateful):
         self.project_name = None
 
         # In multi-worker mode, only the chief of each cluster should report
-        # results.
+        # results and save trials.
         self.multi_worker = False
         self.should_report = True
 
@@ -233,6 +233,8 @@ class Oracle(stateful.Stateful):
                 )
                 trial.metrics.register(metric_name, direction=direction)
             trial.metrics.update(metric_name, metric_value, step=step)
+        if self.should_report:
+            self._save_trial(trial)
         self._save_trial(trial)
         # To signal early stopping, set Trial.status to "STOPPED".
         return trial.status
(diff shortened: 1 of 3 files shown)


keras_tuner/engine/oracle.py
```diff
@@ -233,6 +233,8 @@ def update_trial(self, trial_id, metrics, step=0):
                 )
                 trial.metrics.register(metric_name, direction=direction)
             trial.metrics.update(metric_name, metric_value, step=step)
+        if self.should_report:
+            self._save_trial(trial)
         self._save_trial(trial)
```
haifeng-jin, line 238: Should this line be removed?

haifeng-jin (commented):
Thanks for the PR!

keras_tuner/engine/tuner_utils.py
```diff
@@ -201,6 +206,10 @@ def on_epoch_end(self, epoch, logs=None):
         if self.objective.better_than(current_value, self.best_value):
             self.best_value = current_value
             self.model.save_weights(self.filepath)
+            # Remove temporary saved model files from non-chief workers.
+            ds_utils.remove_temp_dir_with_filepath(
+                self.filepath, self.model.distribute_strategy
+            )
```
haifeng-jin, line 212: Two questions about this part:
1. Why do we only remove the temp dir when we got a better result? Do we do it only when saving weights? However we also did a save_weights on line 203.
2. In the doc of `remove_temp_dir_with_filepath()`, the first arg should be the original file path, but it seems to be using the temporary file path assigned on line 197.

haifeng-jin (commented):
Please also check the linting errors with `sh shell/lint.sh`.
Thanks.

keras_tuner/engine/oracle.py
jamesmullenbach: Yes, thanks!

haifeng-jin (approved):
LGTM

PR #690, opened 2022-05-18
tuner_utils: use datetime to calculate elapsed time

mebeim:
As it currently stands, the elapsed time calculated in `tuner_utils.py` does not take into account durations longer than 24 hours, and wraps around back to "00h 00m 00s" every 24h. It also uses `time.gmtime()` to express the elapsed time, which is conceptually wrong as `time.gmtime()` returns a `struct_time` whose purpose is to represent a *time value*, not a *duration* (delta between two time values).

Use `datetime` instead of `time`, which produces `timedelta` objects that can correctly represent durations, and format the duration accordingly, taking into account the possibility of a total elapsed exceeding 24 hours.

### DIFF
--- a/keras_tuner/engine/tuner_utils.py
+++ b/keras_tuner/engine/tuner_utils.py
@@ -16,7 +16,7 @@
 
 import collections
 import math
-import time
+from datetime import datetime
 
 import numpy as np
 import tensorflow as tf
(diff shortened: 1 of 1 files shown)


haifeng-jin (approved):
Thanks for the PR!

PR #696, opened 2022-06-03
Add "declare_hyperparameters" to HyperModel

jamesmullenbach:
This enables some customization that lets users avoid creating Keras models just to populate the hyperparameter space. This is needed to enable some multi-worker workflows. 

e.g.:
```python
class MyHyperModel(keras_tuner.HyperModel):

  def declare_hyperparameters(self, hp):
    # Declare data processing hyperparameters
    hp.Int('batch_size', 32, 128, step=32, default=64)
    hp.Choice('units', [32, 64, 128])

  def build(self, hp):
    """Builds a convolutional model."""
    inputs = keras.Input(shape=(28, 28, 1))
    x = keras.layers.Flatten()(inputs)
    x = keras.layers.Dense(
        units=hp.get('units'), activation='relu')(
            x)
    outputs = keras.layers.Dense(10)(x)

    model = keras.Model(inputs=inputs, outputs=outputs)
    model.compile(optimizer='adam', loss='mse')
    return model
```

It's very possible there's already a way to do this? Either way, open to suggestions on naming/design.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ You can also check out other versions in our
 Import KerasTuner and TensorFlow:
 
 ```python
-import keras_tuner as kt
+import keras_tuner
 from tensorflow import keras
 ```
 
@@ -66,7 +66,7 @@ We use `objective` to specify the objective to select the best models,
 and we use `max_trials` to specify the number of different models to try.
 
 ```python
-tuner = kt.RandomSearch(
+tuner = keras_tuner.RandomSearch(
     build_model,
     objective='val_loss',
     max_trials=5)
--- a/keras_tuner/__init__.py
+++ b/keras_tuner/__init__.py
@@ -32,4 +32,4 @@ from keras_tuner.utils import check_tf_version
 
 check_tf_version()
 
-__version__ = "1.1.2.master"
+__version__ = "1.1.3dev"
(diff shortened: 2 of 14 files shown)


haifeng-jin (commented):
Seems the logic is: only run `hypermodel.build()` when `declare_hyperparameters()` is not implemented. Would you elaborate more on the use case for this?

Currently, if the user wants to have some data preprocessing hps, they should do them in `HyperModel.fit()`.

jamesmullenbach:
Hyperparameter definitions can't always go in `fit` because for some oracles, all HP's need to be known before setting up some oracles (e.g. Vizier)

For my use case in particular, we want to avoid creating a model, due to the restriction that prevents creating a MultiWorkerMirroredStrategy after other TF ops are run ("Known Issues" [here](https://www.tensorflow.org/api_docs/python/tf/distribute/MultiWorkerMirroredStrategy)). This change would enable us to choose a strategy *after* getting hyperparameters from an oracle. This will enable flexible trial sizing based on chosen hyperparameters. 

In general, calling build() here in general should maybe not be strictly necessary, unless the conditional hyperparameter handling is needed, as it consumes some memory and compute.
An alternate design is to implement `declare_hyperparameters` as just calling `build` in the base `HyperModel` class. Then the only change needed in `BaseTuner._populate_initial_space` is to replace `build` with `declare_hyperparameters`

haifeng-jin (approved):
Let's go with the current implementation for now. It would not affect any current workflows.
We can always change it later if we find a better way.

Issue #372, opened 2020-08-03
Hyperband doesn't use best checkpoints from the previous round


EvsanDlg:
As far as I could understand from the keras-tuner sources, this realization of HyperBand uses the set of best hyperparameters from the previous rounds, but not that models' weights/checkpoints. Models for new rounds initializes with random weights.
Is this behaviour consistent with the original work? Are there some plans to change this behaviour?

Any way, how could one add corresponding checkpoints' loading at the beginning of rounds to the existing workflow of KerasTuner? Any minimal working example would be appreciated!
Thanks!

**UPD.** Ok, I managed to create my own solution by subclassing Hyperband tuner and overwriting `_on_train_begin`, that fires just before `model.fit` in `run_trial` method:
```
from kerastuner.engine import hypermodel as hm_module

class MyTuner(Hyperband):
    def _on_train_begin(self, model, hp, *fit_args, **fit_kwargs):
        prev_trial_id = hp.values['tuner/trial_id'] if 'tuner/trial_id' in hp else None
        if prev_trial_id:
            prev_trial = self.oracle.trials[prev_trial_id]
            best_epoch = prev_trial.best_step
            # the code below is from load_model method of Tuner class
            with hm_module.maybe_distribute(self.distribution_strategy):
                model.load_weights(self._get_checkpoint_fname(
                    prev_trial.trial_id, best_epoch))
```
Check it out. It seens to work fine, as can be seen at tensorflow acc plot:
![image](https://user-images.githubusercontent.com/15376599/89208669-7798d600-d5c5-11ea-9d2f-5da5d1b33568.png)
Any comments are appreciated

yixingfu:
This looks very good!

But I don't think the original [paper for hyperband](https://arxiv.org/pdf/1603.06560.pdf) suggest specifically about whether to use weight from previous round (or maybe I just went through the paper too carelessly)? Or is there any newer work suggesting this approach is better that you can point me to?

In general it may be interesting to include an option to use previous training result for further trials, but it probably not always good (sometimes early trial result may mislead).

EvsanDlg:
@yixingfu , sorry that my answer took so long. I saw the mention of such approach in [Kevin Jamieson's article at Washington University website](https://homes.cs.washington.edu/~jamieson/hyperband.html). It's written that:
[...]

PR #711, opened 2022-08-04
load hyperband model weights from checkpoints when round > 0

Firas-RHIMI:
Since I had the same problem, this commit is related to the problem described in the issue:
https://github.com/keras-team/keras-tuner/issues/372

The current Hyperband tuner implements a method named. "_build_model" to define the behaviour when round_num > 0 in a bracket, and where the model should train from an initial_epoch > 0 (by loading the weights from a saved model). However, this method "_build_model" is not used in the parent class (It should be a naming problem but the method that was defined in the parent class and that is used to build the model is named "_build_hypermodel" . Therefore, the current behaviour does no take into account this logic for Hyperband tuner , and all the training are done from scratch.

Overriding the method "_build_hypermodel" in the Hyperband class with the same logic that was implemented in "_build_model" solves the issue.

### DIFF
--- a/keras_tuner/tuners/hyperband.py
+++ b/keras_tuner/tuners/hyperband.py
@@ -383,15 +383,10 @@ class Hyperband(tuner_module.Tuner):
             fit_kwargs["initial_epoch"] = hp.values["tuner/initial_epoch"]
         return super(Hyperband, self).run_trial(trial, *fit_args, **fit_kwargs)
 
-    def _build_model(self, hp):
-        model = super(Hyperband, self)._build_model(hp)
+    def _build_hypermodel(self, hp):
+        model = super(Hyperband, self)._build_hypermodel(hp)
         if "tuner/trial_id" in hp.values:
             trial_id = hp.values["tuner/trial_id"]
-            history_trial = self.oracle.get_trial(trial_id)
             # Load best checkpoint from this trial.
-            model.load_weights(
-                self._get_checkpoint_fname(
-                    history_trial.trial_id, history_trial.best_step
-                )
-            )
+            model.load_weights(self._get_checkpoint_fname(trial_id))
         return model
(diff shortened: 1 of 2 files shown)


keras_tuner/tuners/hyperband.py
```diff
@@ -383,15 +383,11 @@ def run_trial(self, trial, *fit_args, **fit_kwargs):
...
         if "tuner/trial_id" in hp.values:
             trial_id = hp.values["tuner/trial_id"]
-            history_trial = self.oracle.get_trial(trial_id)
             # Load best checkpoint from this trial.
-            model.load_weights(
-                self._get_checkpoint_fname(
-                    history_trial.trial_id, history_trial.best_step
-                )
-            )
+            print("Load weights from saved model", trial_id)
```
haifeng-jin, line 391: Please remove this line. We will not print information until we have a centralized way to control the verbosity.

haifeng-jin (changes requested):
Thanks for the PR! This really solves an important bug.
It generally looks good except we don't use print.

keras_tuner/tuners/hyperband.py
Firas-RHIMI: Thank you for your review ! I removed the line and formatted the code as stated.

Firas-RHIMI:
Thank you very much for your review ! I removed the print and did the formatting.

haifeng-jin (approved):

Firas-RHIMI:
It seems I still have a problem with the formatting, even though I ran the script ..

haifeng-jin:
@Firas-RHIMI
Please fix the following issue.
`./keras_tuner/tuners/hyperband_test.py:248:86: E501 line too long (91 > 85 characters)`

Firas-RHIMI:
> @Firas-RHIMI Please fix the following issue. `./keras_tuner/tuners/hyperband_test.py:248:86: E501 line too long (91 > 85 characters)`

@haifeng-jin Thank you again for the review and sorry for the multiple messages, I fixed the problem, hope this time the format tests will pass

PR #714, opened 2022-08-11
Add f-string format with flynt

edumucelli:
Hi,

Similar to https://github.com/keras-team/keras/pull/16872 I want to propose to standardize the f-string usage of the whole codebase. I have formated it with [flynt](https://github.com/ikamensh/flynt). Considering the comment from @fchollet on the linked PR, I have not added `flynt` neither to the `shell/format` nor `shell/lint`. The reformatting was all made by flynt. I am not using flynt's --aggressive to avoid "conversions with potentially changed behavior".

### DIFF
--- a/keras_tuner/engine/base_tuner.py
+++ b/keras_tuner/engine/base_tuner.py
@@ -103,9 +103,7 @@ class BaseTuner(stateful.Stateful):
         self._populate_initial_space()
 
         if not overwrite and tf.io.gfile.exists(self._get_tuner_fname()):
-            tf.get_logger().info(
-                "Reloading Tuner from {}".format(self._get_tuner_fname())
-            )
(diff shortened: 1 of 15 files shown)


keras_tuner/engine/base_tuner.py
```diff
@@ -346,9 +344,9 @@ def results_summary(self, num_trials=10):
             num_trials: Optional number of trials to display. Defaults to 10.
         """
         print("Results summary")
-        print("Results in %s" % self.project_dir)
+        print(f"Results in {self.project_dir}")
         print("Showing %d best trials" % num_trials)
```
haifeng-jin, line 348: Why this is not changed?

haifeng-jin (approved):
Thanks for the PR!

PR #741, opened 2022-10-11
Add comments for backward compatibility

haifeng-jin:


haifeng-jin (commented):
Add comments for backward compatibility

Issue #704, opened 2022-07-06
A standard way to report more failure status of a trial


haifeng-jin:
**Is your feature request related to a problem? Please describe.**

If a model is failed at training, the whole program crashes.
A better strategy is to report the status fo this trial and skip it.

**Describe the solution you'd like**

One solution would be have a new function of Oracle just to report the failure status of a trial.
However, it requries modification to the proto files used by gRPC.
Rurther looks into the roadmap of distributed features of keras_tuner is required.

**Describe alternatives you've considered**

**Additional context**

carlthome:
This is sorely missing.

I'm tuning both feature representation and model topology simultaneously in a keras-tuner experiment, and when the number of pooling layers is too many for the size of the input features, the whole tuning simply crashes, with no obvious way for me to catch exceptions and skip to the next trial.

haifeng-jin:
This can now be done with `raise keras_tuner.FailedTrialError("error message.")` in `HyperModel.build()`, `HyperModel.fit()`, or your model build function.

PR #757, opened 2022-11-16
split trial status to oracle status

haifeng-jin:
This is preparing for #704.

Split `TrialStatus` into `TrialStatus` and `OracleStatus`.
`TrialStatus` was used in `Trial.status`, which is abused for several different usages, which will add a lot of complexity to the fault tolerance support if not fixed.
It was used for:
* Used by `Oracle.create_trial()` to indicate the `Oracle` status. Indicate if an oracle is ready for creating another trial, or waiting for results of other trials, or has stopped for some reason. This information is used by the tuner to decide whether to create a trial.
* After a trial is finished it is used to mark the trial result as invalid or completed normally.
* It is also used by the oracle to inform a trial to stop when calling `Oracle.update_trial()`.

To make the logic clearer, we have split the first usage above to `OracleStatus`, and keep the rest two still in `TrialStatus`.
`Oracle.create_trial()` is not using `Trial.status` to indicate the `Oracle` status anymore.
It now returns two values: `(OracleStatus, Trial)`.
We do not expect users to override `Oracle.create_trial()`. None of our current tuning algorithms override that function.
So it would not break the users.


haifeng-jin:
This doesn't seem to be a good design due to the extra status value returned by create_trial.
Created another pull request #760

PR #775, opened 2022-12-04
add tensorflow as a dependency

haifeng-jin:
KerasTuner cannot be imported without TensorFlow installed. So adding
TensorFlow as a dependency. The reason why it was not added earlier is that we
want the user to choose between tensorflow-cpu and tensorflow-gpu, which is no
longer a problem.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -30,7 +30,6 @@ jobs:
             ${{ runner.os }}-pip-
       - name: Install dependencies
         run: |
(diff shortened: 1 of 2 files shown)


MartinNowak:
> The reason why it was not added earlier is that we
> want the user to choose between tensorflow-cpu and tensorflow-gpu, which is no
> longer a problem.

Well it is a problem that we can no longer use tensorflow-cpu which is only 200MiB vs. 550MiB.

PR #789, opened 2022-12-21
change how we check method overriden

haifeng-jin:
The try except clause would add extra info the stacktrace if exception happens
in the except block. Change it to check override methods directly.

cc: @jamesmullenbach

### DIFF
--- a/keras_tuner/engine/base_tuner.py
+++ b/keras_tuner/engine/base_tuner.py
@@ -128,6 +128,34 @@ class BaseTuner(stateful.Stateful):
             # Only populate initial space if not reloading.
             self._populate_initial_space()
 
+    def _activate_all_conditions(self, hp):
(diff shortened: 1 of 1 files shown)


keras_tuner/engine/base_tuner.py
```diff
@@ -128,6 +128,34 @@ def __init__(
...
 
+    def _activate_all_conditions(self, hp):
+        # Lists of stacks of conditions used during `explore_space()`.
+        scopes_never_active = []
+        scopes_once_active = []
+
+        while True:
+            self.hypermodel.build(hp)
+
+            # Update the recored scopes.
```
jamesmullenbach, line 139: Should "recored" be "recorded"?

jamesmullenbach (approved):
LGTM!

PR #794, opened 2022-12-22
Fix default `num_initial_points` to match doc

leleogere:
Currently, the argument `num_initial_points` of the `BayesianOptimization` tuner defaults to 2 even if the doc states that it should default to 3*dimension when unspecified.

I just made it default to `None` in order to avoid having to pass `None` manually to get the behaviour specified in the doc.


haifeng-jin (approved):
Thanks for the PR!

Issue #766, opened 2022-11-28
Partially support of `Jax` for Bayesian opt


Anselmoo:
**Is your feature request related to a problem? Please describe.**

Optimisation via `Bayesian` might cause some performance issues due to the evaluation of the _Einstein Summation_.

https://github.com/keras-team/keras-tuner/blob/a7a361f9521cb1033a05aba865c86eb30784d907/keras_tuner/tuners/bayesian.py#L124-L127

**Describe the solution you'd like**

Let performing [`jax.numpy.einsum`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.einsum.html) can help.

Partially support can look like:

```python
try:
    from jax.numpy import einsum
except ImportError:
    from numpy import einsum
```

or system dependent:

```python
if sys.platform != "win32":
    import jax.numpy as np
else:
    import numpy as np
```

Anselmoo:
@haifeng-jin, thx for the feedback; that's an excellent idea. I will setup a repo for benchmark!

Anselmoo:
Dear @haifeng-jin

To be honest, I don't see a performance difference at all based on my manual test. 

It's tested with the same test as [`bayesian_test.py`](https://github.com/Anselmoo/keras-tuner/blob/Anselmoo/issue766/keras_tuner/tuners/bayesian_test_tf.py), and I tried to use [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/), but it didn't show any useful data.

So, I set up a local `time` based and got the following results for `numpy` vs. `tf.experimental`:

* Total time for Numpy Oracle:  727.1199822425842 sec for three runs
* Total time for TensorFlow Oracle:  780.9310626983643 sec for three runs

The log files are [here](https://github.com/Anselmoo/keras-tuner/blob/Anselmoo/issue766/keras_tuner/tuners/benchmark/results.log)

It's noticeable that after 60 trials, the `total elapsed time` is a little bit fluctuating between seconds and minutes. Maybe it's related to how the new hyper parameters will be added? 



### About the implementation

```python
  y_var = np.ones(len(x), dtype=np.float)
        y_var -= tf.experimental.numpy.einsum(
        y_var -= tf.experimental.numpy.einsum(
            "ij,ij->i",
            tf.experimental.numpy.dot(kernel_trans, kernel_inv),
            kernel_trans,
        )

        # Convert y_var[y_var < 0] = 0.0 into a tensorflow conform way.
[...]

PR #795, opened 2022-12-22
feat: :sparkles: Repalce of `Numpy` routines via `TensorFlow`

Anselmoo:
### Scope:

According to issue #766 and the comment about using [TensorFlow-Numpy](https://www.tensorflow.org/api_docs/python/tf/experimental/numpy), a partially refactor via `Tensorflow-Routines`of the bayesian module is provided.

### Comment:

Due to the computing cost of the diagonalization methods, a replacement of this module's `NumPy` and `SciPy` routines might increase the performance. 

See also [tfp.optimizer.lbfgs_minimize](https://www.tensorflow.org/probability/api_docs/python/tfp/optimizer/lbfgs_minimize) for `scipy.optimize.minimize`.

Fixes: #766

### DIFF
--- a/keras_tuner/tuners/bayesian.py
+++ b/keras_tuner/tuners/bayesian.py
@@ -33,7 +33,7 @@ from keras_tuner.engine import tuner as tuner_module
 def cdist(x, y=None):
     if y is None:
         y = x
-    return np.linalg.norm(x[:, None, :] - y[None, :, :], axis=-1)
+    return tf.linalg.norm(x[:, None, :] - y[None, :, :], axis=-1).numpy()
 
 
 def solve_triangular(a, b, lower):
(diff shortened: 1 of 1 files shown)


haifeng-jin:
@Anselmoo Thanks for the PR!
Unfortunately, we decided to change back to the sklearn implementation to reduce the maintenance burden for now as we are supporting parallel tuning. This is a temporary regression. See #798.

Another note is that I tried TF implementation in #506, which is much slower than the scipy implementation.

We will refer to this in the future when we migrate to the TF implementation. We will do that when TF has proved to be faster than the sklearn implementation.

Anselmoo:
@haifeng-jin thx for the feedback and no worries.

There is no doubt that `sklearn` implementations are powerful and established routines. 

Next to `JAX`, TF implementation for NumPy looks like a big benefit for other numerical challenges like _newton raphson methods_ or ... thx for mentioning (https://github.com/keras-team/keras-tuner/issues/766#issuecomment-1341258762) wasn't aware of it until now!

PR #806, opened 2023-01-06
Change `end_trial` signature

haifeng-jin:
Some newly appeared `HyperParameter`s during a `Trial` needs to be write back to the corresponding `Trial` in `Oracle`.
It is fine for single machine settings, where the same `Trial` object are shared between the `Oracle` and the `Tuner`.
However, for parallel tuning, the `Trial` are passed to the workers through gRPC. The new `HyperParameter`s appeared on the worker side are not updated on the `Oracle` side.

We use `Oracle.update_space(hp)` to update the newly appeared `HyperParameter`s before this PR, which does only accepts a `HyperParameters` object, no `trial_id` to identify the corresponding `Trial`.

We propose changing the current signature of `Oracle.end_trial(trial_id, status, message)` to `Oracle.end_trial(trial)`. The `trial` argument contains all the information in old signature arguments and contains the `HyperParameters` object, which contains the newly appeared `HyperParameter`s we need to write back.

It is more consistent with `Oracle.create_trial()` logically. `Oracle.create_trial()` returns a `Trial`. After the trial, the updated `Trial` is reported back with `Oracle.end_trial(Trial)`.

In this way, we also do not need to call `Oracle.update_space()` in the `Tuner` class since it can be done during the `Oracle.end_trial()`.

It also makes the code cleaner, from:

https://github.com/keras-team/keras-tuner/blob/aec47050039483c78ec2cc6c065b4e20ef68e76b/keras_tuner/engine/base_tuner.py#L319-L320

to:

```py
 self.oracle.end_trial(trial)
```

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -5,8 +5,8 @@
   with `Oracle.update_trial()` in `Tuner.run_trial()` is deprecated. Please
   return the metrics in `Tuner.run_trial()` instead.
 * If you implemented your own `Oracle` and overrided `Oracle.end_trial()`, you
-  need to change the signature of the function to
-  `Oracle.end_trial(trial_id, status, message)`.
+  need to change the signature of the function from
+  `Oracle.end_trial(trial.trial_id, trial.status)` to `Oracle.end_trial(trial)`.
 * The default value of the `step` argument in `keras_tuner.HyperParameters.Int()` is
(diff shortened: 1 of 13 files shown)


haifeng-jin:
@jamesmullenbach Would you help confirm if this would impact AutoDistributionTuner somehow?

PR #816, opened 2023-01-12
Resolves #813

### LINKED ISSUE #813: Change all Py2 style super() class call to Py3 style
Change all `super(ClassName, self).method_name()` occurance in the codebase to `super().method_name()`, which is recommended by the Python 3 documentation.

[More details](https://stackoverflow.com/a/10483204/7910419).

For example, this:

https://github.com/keras-team/keras-tuner/blob/d536635a92ca4c4177d4d77f2ee893b301ed6090/keras_tuner/engine/oracle.py#L482

needs to be changed to:

```py
super().reload(self._get_oracle_fname())
```

Please comment on this issue before you start working on it so that others can know.

LuNoX:
Resolves #813

All super() class call styles have been updated to Python3 style. 
Previously, all super() classes were called in a Python2 style passing the inheriting class and self as arguments to super(). This is no longer recommended with Python3. 
As a result, all super() class calls were changed to Python3 style.

### DIFF
--- a/keras_tuner/applications/augment.py
+++ b/keras_tuner/applications/augment.py
@@ -178,7 +178,7 @@ class HyperImageAugment(hypermodel.HyperModel):
             # `randaug_count` is set to 0.
             self.model_name = "image_augment"
 
-        super(HyperImageAugment, self).__init__(**kwargs)
+        super().__init__(**kwargs)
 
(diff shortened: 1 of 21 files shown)


haifeng-jin (approved):
Thanks for the PR!

PR #834, opened 2023-02-08
Preparation for switch in Keras serialization format

nkovela1:
This PR changes serialization API calls in KerasTuner to the legacy APIs in preparation for a switch to the new Keras object serialization format. There is no change in functionality now by this switch to the legacy APIs.


haifeng-jin (approved):
Thanks for the PR!

Issue #839, opened 2023-02-10
Add new extra require for bayesian


Anselmoo:
**Is your feature request related to a problem? Please describe.**

Currently, `sklearn` and `scipy` are part of the extras test.

https://github.com/keras-team/keras-tuner/blob/aba01ea8dc7eb6040451a71e1be123db868b8954/setup.py#L46-L47

Due to the importance of the [Bayesian-Opt](https://github.com/keras-team/keras-tuner/blob/master/keras_tuner/tuners/bayesian.py) it might make sense to give them an _own_ extra. 

**Describe the solution you'd like**

```py
extras_require={
        "tests": [
            "black",
            "flake8",
            "isort",
            "pandas",
            "portpicker",
            "pytest",
            "pytest-cov",
            "pytest-xdist",
            "namex",
         ],
       "bayesian": [
            "scikit-learn",
            "scipy",
        ],

```

PR #843, opened 2023-02-16
build: :package: Add extra require for `bayesian`

Anselmoo:
Fixes: #839

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,13 @@ Install the latest release:
 pip install keras-tuner --upgrade
 ```
 
+Install _optional_ dependencies for the `Bayesian` tuner:
+
+```
+pip install keras-tuner[bayesian]
+```
+
+
 You can also check out other versions in our
 [GitHub repository](https://github.com/keras-team/keras-tuner).
 
--- a/README.md
+++ b/README.md
@@ -31,13 +31,6 @@ Install the latest release:
 pip install keras-tuner --upgrade
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -31,6 +31,13 @@ Install the latest release:
 pip install keras-tuner --upgrade
 ```
 
+Install _optional_ dependencies for the `Bayesian` tuner:
+
+```
+pip install keras-tuner[bayesian]
+```
+
+
```
haifeng-jin, line 40: We would like to keep the instructions in README.md simple and only inform the users in the error messages when they use the `BayesianOptimization`.
So please remove the changes here.

haifeng-jin (changes requested):
Thanks for the PR!

Would you please also help update the error messages in this PR?

https://github.com/keras-team/keras-tuner/blob/8a0f9a7ab5c2d9cdd9ef8cb6f47ea503e17781ed/keras_tuner/tuners/bayesian.py#L102-L111

like
```py
if scipy is None or sklearn is None:
    raise ImportError("Please install the required dependencies for `BayesianOptimization` with `pip install keras-tuner[bayesian]`.")
```

README.md
Anselmoo: Thx @haifeng-jin, I like your idea 😊

Anselmoo: ✅

haifeng-jin (approved):
Thanks!

PR #845, opened 2023-02-16
update how build.py export keras_tuner.__version__

haifeng-jin:
* Updated how `build.py` export `keras_tuner.__version__`.
* Removed the deprecated author email address in `setup.py`.
* Extract the version variable in `setup.py` to make it easier to find.


fchollet (approved):
LGTM. If you want to make it single source of truth, you can also use tags in setup.py that you fill with the correct version string when you copy setup.py as part of the build process.

PR #849, opened 2023-02-22
remove old confusing names of oracles

haifeng-jin:
This PR removes some confusing deprecated import names used before 1.0.2.

* Removed `keras_tuner.oracles.BayesianOptimization`, `keras_tuner.oracles.Hyperband`, `keras_tuner.oracles.RandomSearch`, which were actually `Oracle`s instead of `Tuner`s. Please use`keras_tuner.oracles.BayesianOptimizationOracle`, `keras_tuner.oracles.HyperbandOracle`, `keras_tuner.oracles.RandomSearchOracle` instead.
* Removed `keras_tuner.Sklearn`. Please use `keras_tuner.SklearnTuner` instead.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,6 +3,12 @@
 ## Breaking changes
 * Removed `Logger` and `CloudLogger` and the related arguments in
   `BaseTuner.__init__(logger=...)`.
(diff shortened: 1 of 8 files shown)


fchollet (approved):
LGTM, thanks. This is better.

PR #891, opened 2023-04-04
separate extras for tensorflow-cpu

MartinNowak:
- allow to install `keras-tuner[tensorflow-cpu]`

Maybe a viable alternative to #775?

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -27,6 +27,6 @@ jobs:
           key: ${{ runner.os }}-pip-${{ hashFiles('setup.py') }}
       - name: Install dependencies
         run: |
-          pip install -e ".[tests]" --progress-bar off --upgrade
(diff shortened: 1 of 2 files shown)


haifeng-jin (commented):
Thanks for the PR! This makes sense. I will add a check to catch the import error if TF is not installed by pull request to your branch and we can merge it altogether.

haifeng-jin:
To clarify the necessity for this PR for future reference.

People may want to use `tensorflow-cpu` instead of `tensorflow` with KerasTuner.
Especially, when they use KerasTuner to tune non-TensorFlow related things.
We don't want to force install tensorflow in these cases.

For people who, want to use KerasTuner with tensorflow with GPU, they should have installed it already with complicated setups. So usually not even need it from KerasTuner dependency.

MartinNowak:
> Thanks for the PR! This makes sense. I will add a check to catch the import error if TF is not installed by pull request to your branch and we can merge it altogether.

Added a commit to the CI test @haifeng-jin.

PR #930, opened 2023-09-12
Set workflow permissions to read-only

pnacht:
Fixes https://github.com/keras-team/keras-tuner/issues/929.

This PR ensures all workflows run with read-only permissions, protecting the project from supply-chain attacks.


haifeng-jin (approved):
Thanks for the patch!

PR #977, opened 2023-11-15
save chief's state before starting oracle server

### LINKED ISSUE #976: Chief process does not resume tuning after restart due to missing `chief.json` file
**Describe the bug**
In a [distributed tuning](https://keras.io/guides/keras_tuner/distributed_tuning/) setting, in case the chief process is discontinued and after restarting it, it starts with trial `#1` again instead of resuming at the last unfinished trial. 

After some investigation, I found out that the `BaseTuner` class checks for a tuner-specific file to exist (in this case, it's `chief.json`):
https://github.com/keras-team/keras-tuner/blob/e935ac15d72a2b183b79e8c99bed49f91708681c/keras_tuner/engine/base_tuner.py#L126-L129

However, this file is only created for worker processes, not for the chief process, so for the latter the condition will always be `false` and cause it to not reload the state and resume tuning. 

Creating the file manually makes the chief process to continue, as expected.

**To Reproduce**
Use the code from [the documentation](https://keras.io/guides/keras_tuner/distributed_tuning/) but with `RandomSearch` instead of `Hyperband` to ensure long running tuning and setting parameter `overwrite=False`.

In two terminals, run the chief and worker processes as described in the documentation, then after a while, kill both processes and run them again.

**Expected behavior**
The chief process should print `Reloading Tuner from results_dir/mnist/chief.json` and resume with the last running trial.

**Additional context**
Python 3.10.13
keras-tuner 1.4.6

**Would you like to help us fix it?**
I'd like too, but time is short, so I may need help to point me at the right location in the code to look after.

sfo:
This fixes #976 by saving the chief process' state before starting the oracle server. So when running the chief with `overwrite=False`, it gets reloaded properly and resumes tuning with the most recent trial.

### DIFF
--- a/keras_tuner/engine/base_tuner.py
+++ b/keras_tuner/engine/base_tuner.py
@@ -215,6 +215,7 @@ class BaseTuner(stateful.Stateful):
             # Avoid import at the top, to avoid inconsistent protobuf versions.
             from keras_tuner.distribute import oracle_chief
 
+            self.save()
             oracle_chief.start_server(self.oracle)
             return
 


haifeng-jin (approved):
Great work! Thanks for the PR!

Issue #990, opened 2024-03-04
Distributed keras tuner fails in case when chief oracle server startup takes more than 5 minutes


onponomarev:
We observed that in some cases distributed keras tuner fails.

It's caused by the fix to prevent keras tuner hanging forever - 5 min RPC timeouts were introduced (see #957). 
Now if keras chief worker startup takes longer than 5 minutes, then the client gives up and fails the whole tuning process.

Normally RPC server startup is quick, but in some cases it might take slightly longer. 

The planned solution is to increase the client timeout to 1h. We still need the timeout to prevent tuner clients from hanging forever. We need the timeout to be high enough so that chief oracle server always has enough time to start.

PR #991, opened 2024-03-04
Increase oracle client timeout from 5 to 60 minutes.

onponomarev:
The timeout is so high to prevent a rare race condition from happening. We need clients to wait till chief oracle server starts. This normally takes a few minutes, but sometimes might take longer.
See https://github.com/keras-team/keras-tuner/issues/990 for more details. Initially we didn't have any timeout. It was introduced to avoid tuner jobs hanging forever if chief oracle stops responding.
See https://github.com/keras-team/keras-tuner/pull/957.

### DIFF
--- a/keras_tuner/distribute/oracle_client.py
+++ b/keras_tuner/distribute/oracle_client.py
@@ -21,7 +21,14 @@ from keras_tuner import protos
 from keras_tuner.engine import hyperparameters as hp_module
 from keras_tuner.engine import trial as trial_module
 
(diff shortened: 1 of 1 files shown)


haifeng-jin (approved):
Thanks for the PR!

haifeng-jin:
A few more  changes needed. I will take over.
Merging it into temp_fix branch to do the fixes.
I will then merge temp_fix into master and do the release.

PR #1059, opened 2026-06-14
Document pickle trust boundary in SklearnTuner model loading

AgentGymLeader:
## What

Add explicit security documentation and a soft runtime warning for the
pickle trust boundary in `SklearnTuner`.

Changes are limited to `keras_tuner/tuners/sklearn_tuner.py`:

- **`import warnings`** (stdlib, no new dependency).
- **Class docstring** — new *Security note* section explaining that
  fitted models are persisted as `model.pickle` files and deserialized
  via `pickle.load` when `get_best_models()` is called or a search is
  resumed with `overwrite=False`; the project `directory` must therefore
  be trusted and not shared-writable. References the scikit-learn
  model-persistence docs and notes the contrast with the standard
  `keras_tuner.Tuner`, which deliberately uses `model.load_weights()`
  to avoid full deserialization.
- **`save_model` docstring** — new, restates the trust requirement.
- **`load_model` docstring** — new, restates the trust requirement and
  documents args/returns.
- **`warnings.warn(...)` in `load_model`** — one `UserWarning` emitted
  immediately before `pickle.load`, so callers that reach model loading
  programmatically (e.g. via `get_best_models()`) see the warning at
  runtime.

No signatures, defaults, or runtime behaviour are changed. This is a
documentation + soft-warning hardening only.

## Why

`SklearnTuner.load_model` calls `pickle.load` on a per-trial
`model.pickle` file from the project directory with no warning anywhere
— not in the class docstring, not in the method docstring, not at
runtime. The reachable paths include:

- `tuner.get_best_models()` (the canonical post-search API)
- Automatic project reload when `overwrite=False` (the default)

Pickle deserialization can execute arbitrary code. Loading or resuming a
project directory from an untrusted or shared-writable source therefore
exposes callers to arbitrary-code execution with no prior indication
that deserialization is happening.

The standard `keras_tuner.Tuner.load_model` already handles this
correctly: it uses `model.load_weights()` and includes an explicit
comment explaining why it avoids loading `custom_objects` (full
deserialization). The sklearn path had no equivalent safeguard, creating
an undocumented inconsistency in the library's security posture.

This PR makes the trust assumption explicit and audible without changing
any API or default behaviour.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/keras_tuner/tuners/sklearn_tuner.py
+++ b/keras_tuner/tuners/sklearn_tuner.py
@@ -16,6 +16,7 @@ import collections
 import inspect
 import os
 import pickle
+import warnings
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


AgentGymLeader:
Closing this. It's a documentation-only change, and whether to surface the SklearnTuner pickle trust boundary is better left to the maintainers' own security guidance than added as an unsolicited docstring note. Apologies for the noise.

